PR #16, opened 2016-03-03
Improve historic hook callbacks

vodik:
Historic callbacks always immediately invoke all registered hooks, but the callback is **only** triggered on hooks that come from newly registered plugins. This change captures all information.

The added test fails on the current pluggy, so should catch any regressions should this be accepted.

I found this toying with pytest. The only historic hook in pytest is `pytest_namespace`, and the only reason this issue isn't a problem in pytest is the order in which things are setup. However if someone wanted to implement something similar inside a 3rd party plugin, this inconsistency means some results may be lost even though all hooks get triggered.

### DIFF
--- a/pluggy.py
+++ b/pluggy.py
@@ -747,7 +747,9 @@ class _HookCaller(object):
     def call_historic(self, proc=None, kwargs=None):
         self._call_history.append((kwargs or {}, proc))
         # historizing hooks don't return results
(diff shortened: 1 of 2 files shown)


hpk42:
thanks for the PR.  I am fixing it some more in https://github.com/pytest-dev/pluggy/pull/29

vodik:
Nice, thanks!

PR #21, opened 2016-09-04
Added required argument to HookspecMarker

MrGreenTea:
If this is set to True calls to the hook will raise HookCallError if there is not at least one registered implementation.
It uses `self.specopts.get` with a default value so it stays compatible to manual setting of the specopts dict.

see Issue #20

### DIFF
--- a/pluggy.py
+++ b/pluggy.py
@@ -99,6 +99,7 @@ class HookspecMarker:
         If historic is True calls to a hook will be memorized and replayed
         on later registered plugins.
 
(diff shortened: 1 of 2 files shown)


hpk42:
I am skeptical of adding something to hookspec -- see also on my issue comments.  Why is it not sufficient to add a runtime-api like ``has_implementations()`` to hooks?

hpk42:
As per discussion on #20 i am closing this.

PR #43, opened 2017-02-13
Support default via kwargs

tgoodlet:
NOTE: this PR builds upon #42 and thus contains its changes as well (for now) but will be rebased after that is merged.

This PR is an initial working implementation what was originally requested by @RonnyPfannschmidt in #15.

I still have concerns with this work (namely the reservations some have had with calling hook functions using `f(**kwargs)` due to performance) because I do not believe we have properly benchmarked (looked at average versus best case perf) to discard this alternative approach. The use of `f(**kwargs)` will greatly simplify call loop logic in `_MultiCall.execute()` which I think is worth acknowledging. I have created #37 to fully assess this performance question.

@nicoddemus @hpk42 I realize #15 is quite lengthy but I believe this work should give us at the least some concrete code and a test to help us agree whether this feature is even something desirable considering its somewhat confusing default arguments precedence order.

As a refresher @RonnyPfannschmidt requested that `hookspec` keyword argument declarations would be used as defaults to `hookimpl`s which require them when those arguments are **not** provided at hook call time. Additionally, a `hookimpl` can also define `kwarg`s with default values which will be used in the case that an older version of the `hookspec` defining project is being used and thus no defaults are defined on the `hookspec`. @hpk42 later suggested that such `kwargs` should **only** have a default value of `None` which @RonnyPfannschmidt agreed with.


testing/test_hookrelay.py
```diff
@@ -76,3 +76,32 @@ def hello(self, arg):
...
+
+    class Plugin:
+        @hookimpl
+        def myhook(self, arg, kwarg="my default"):
+            return kwarg
+
+    pm.register(Plugin())
+
+    # with no spec registered
+    assert pm.hook.myhook(arg='yeah!')[0] == "my default"
```
nicoddemus: Hmm I'm surprised by this behavior; I would not expect a hook caller to be able to make a call without a registered spec. Is this the bug @hpk42 commented [here](https://github.com/pytest-dev/pluggy/issues/15#issuecomment-261485446)?

testing/test_hookrelay.py
```diff
@@ -76,3 +76,32 @@ def hello(self, arg):
...
+def test_defaults(pm):
+    """Verify that default keyword arguments can be declared on both specs
+    and impls. The default value look up precedence is up as follows:
+        - caller provided value
+        - hookspec default
+        - hookimpl default
+    """
+    class Api:
+        @hookspec
+        def myhook(self, arg, kwarg="default"):
```
nicoddemus: Looking at this test I kind agree with the idea of only allowing `None` as defaults in specs and impls. This is would be less confusing to users and doesn't prevent us to actually allow other values in the future if one of them becomes "obvious".

testing/test_hookrelay.py
tgoodlet: @nicoddemus no this is expected if you look at the code. If you want to raise an error due to missing `hookspecs` then use [`PluginManager.check_pending()`](http://pluggy.readthedocs.io/en/latest/#enforcing-spec-validation).

testing/test_hookrelay.py
tgoodlet: @nicoddemus yes this was the consensus prior as well.

My main concern with this is we may end up with lots of `if arg is None:` checks throughout `hookimpl`s. I guess it serves as an explicit way to show that your `hookimpl` may be further ahead then the defining project's `hookspec`?

pluggy.py
```diff
@@ -749,10 +761,28 @@ def _maybe_apply_history(self, method):
                     proc(res[0])
 
 
+class HookSpec(object):
```
tgoodlet: Even if we decide to abandon this defaults idea I **really** think we should keep this encapsulation of `hookspec` objects with a type much like we do with `HookImpl`. It makes introspection tasks down the road that much simpler.

tgoodlet:
@RonnyPfannschmidt @nicoddemus @hpk42 alright now that #42 is in this PR is a lot easier to understand and get back to.

The one outstanding that everyone seemed to agree on is that if we decide to allow default arguments in `hookspec`  and `hookimpl` then they should only be allowed to have a `None` default value.

@nicoddemus also clarified terminology nicely in review of #42:
>Let's see if we are using the same nomenclature first. Considering this function:
```python
def foobar(x, y, z=4):
    pass  
```
> - positional arguments: caller passes by position so order matters: foobar(1, 2, 10)
>- keyword arguments: caller passes using keyword syntax so order does not matter: foobar(y=2, x=1, z=10)
>- default arguments: caller does not need to pass them on: foobar(1, 2) # z = 4

On a further note, if we decide this feature is no longer of interest I'd still like to keep a couple things from this PR that are (imho) good improvements. Namely,
- encapsulating `hookspec` objects using a `HookSpec` type
  - makes future introspection tasks much easier
- avoid monkey patching (eg. `_HookCaller._specmodule_or_class`)
  - easier to understand code
- make `_MultiCall` as hook agnostic as possible
  - maintain a separation of concerns

I would like to get some work on #37 in before moving further with changes here but I want to get everyone thinking about it nonetheless.

Thanks!

tgoodlet:
@RonnyPfannschmidt I'm guess we can close this one then?

Issue #44, opened 2017-02-16
Move to new style classes


tgoodlet:
As per pytest-dev/pytest#2147 we should also move to new-style classes throughout!

tgoodlet:
Closing as this was completed in #44

The-Compiler:
You mean #46, this is #44 :wink:

tgoodlet:
Hah (facepalm)

PR #46, opened 2017-02-16
Moving to new style classes (object inheritance) - Issue #44

MichalTHEDUDE:
Adding to all classes inheritance from `object` class.

All test green, excepts for Interpreters:
- python2.6
- python3.4
- pypy

as I don't have them. 
Hope it isn't an issue :)

### DIFF
--- a/docs/examples/firstexample.py
+++ b/docs/examples/firstexample.py
@@ -4,7 +4,7 @@ hookspec = pluggy.HookspecMarker("myproject")
 hookimpl = pluggy.HookimplMarker("myproject")
 
 
-class MySpec:
+class MySpec(object):
     """A hook specification namespace.
     """
     @hookspec
@@ -13,7 +13,7 @@ class MySpec:
         """
 
 
-class Plugin_1:
+class Plugin_1(object):
     """A hook implementation namespace.
     """
     @hookimpl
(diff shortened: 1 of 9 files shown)


nicoddemus:
Not at all, we have AppVeyor and Travis to test across all interpreters and configurations required. Thanks again!

hpk42:
Is there any particular reason to move to new style classes? I typically try to use a subset of python such that it doesn't matter if a class derives from object.  In fact i rarely use inheritance anyway :)

nicoddemus:
This has been discussed in #2147, old-style classes introduce subtle differences that, granted, are hard to come by, but since it is simple to make it consistent across all versions we decided to just make them all new-style explicitly.

tgoodlet:
@hpk42 @nicoddemus it's an interesting point too because [in py3 old-style don't exist any more](https://docs.python.org/2/reference/datamodel.html#new-style-and-classic-classes) and [the syntax is equivalent](https://docs.python.org/3.6/reference/compound_stmts.html#class).

hpk42:
oki.  could you directly commit a changelog entry?

tgoodlet:
oh yey and now I have to rebase everything lolz

PR #74, opened 2017-08-30
Update version to 0.5.2.dev

nicoddemus:
This helps to differentiate from an official release and dev version.

Until #73 is done we have to remember to do that manually after the release as well.


tgoodlet:
Good call @nicoddemus. 
I totally forgot `setuptools` supports this.

PR #77, opened 2017-08-30
Fix error while issuing DeprecationWarning for __multicall__ usage

### LINKED ISSUE #75: Error: warnings has no attribute DeprecationWarning
While testing `pluggy` off the branch from #72, I came across this error:

```
        if '__multicall__' in hookimpl.argnames:
            warnings.warn(
                "Support for __multicall__ is now deprecated and will be"
                "removed in an upcoming release.",
>               warnings.DeprecationWarning
            )
E           AttributeError: module 'warnings' has no attribute 'DeprecationWarning'
```

https://github.com/pytest-dev/pluggy/blob/557cd3f0245347c1cd359c6deb4b25cde944e98a/pluggy/__init__.py#L656-L661

And indeed `warnings` module doesn't have a `DeprecationWarning`.

We should add a test to catch this warnings (and any others we might have).

nicoddemus:
Fix #75

### DIFF
--- a/pluggy/__init__.py
+++ b/pluggy/__init__.py
@@ -657,7 +657,7 @@ class _HookCaller(object):
             warnings.warn(
                 "Support for __multicall__ is now deprecated and will be"
                 "removed in an upcoming release.",
-                warnings.DeprecationWarning
+                DeprecationWarning
             )
             self.multicall = _LegacyMultiCall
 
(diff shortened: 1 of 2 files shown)


nicoddemus:
Test failed because I used the context manager version of `pytest.deprecated_call` in my test, and that is not available in older versions.

But I just realized, what's the point of running the `pluggy` test suite using older pytest versions? Up until now `pytest` vendored pluggy anyway.

tgoodlet:
@nicoddemus i'm cool with always using latest `pytest` 👍

PR #80, opened 2017-09-01
Add some __tracebackhide__ statements expected by pytest

nicoddemus:
This regression was discovered by pytest-dev/pytest#2744. 😅


nicoddemus:
I will just go ahead and merge this given that it is trivial and will allow us to go ahead with pytest-dev/pytest#2744.

Issue #64, opened 2017-08-24
Document how hook call order is affected by plugin registration


nicoddemus:
As discussed in #62, we should explicitly document how hook call order is done regarding how plugins are registered. 

From the top of my head, hooks are called in this order:

1. setuptools plugins (in arbitrary order, except those marked with `try_first` and `try_last`);
2. core plugins;

astrojuanlu:
> setuptools plugins (in arbitrary order

I know this is an ancient issue but I don't think this fact was sufficiently documented in gh-85.

PR #85, opened 2017-09-07
Some outstanding docs

tgoodlet:
Resolves a couple docs issues (#49, #64) and adds a new test for `hookimpl` order verification.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -152,9 +152,10 @@ then the *hookimpl* should be marked with the ``"optionalhook"`` option:
 
 Call time order
 ^^^^^^^^^^^^^^^
-A *hookimpl* can influence its call-time invocation position.
-If marked with a ``"tryfirst"`` or ``"trylast"`` option it will be
-executed *first* or *last* respectively in the hook call loop:
+By default hooks are :ref:`called <calling>` in LIFO registered order, however,
+a *hookimpl* can influence its call-time invocation position using special
+attributes. If marked with a ``"tryfirst"`` or ``"trylast"`` option it
+will be executed *first* or *last* respectively in the hook call loop:
 
 .. code-block:: python
 
@@ -477,6 +478,8 @@ You can retrieve the *options* applied to a particular
     http://doc.pytest.org/en/latest/writing_plugins.html#setuptools-entry-points
 
 
+.. _calling:
+
 Calling Hooks
 *************
 The core functionality of ``pluggy`` enables an extension provider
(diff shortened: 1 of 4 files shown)


docs/index.rst
```diff
@@ -152,9 +152,10 @@ then the *hookimpl* should be marked with the ``"optionalhook"`` option:
 
 Call time order
 ^^^^^^^^^^^^^^^
-A *hookimpl* can influence its call-time invocation position.
-If marked with a ``"tryfirst"`` or ``"trylast"`` option it will be
-executed *first* or *last* respectively in the hook call loop:
+By default hooks are :ref:`called <calling>` in LIFO registered order, however,
+a *hookimpl* can influence its call-time invocation position using special
+attributes. If marked with a ``"tryfirst"`` or ``"trylast"`` option it
+will be executed *first* or *last* respectively in the hook call loop:
```
nicoddemus: What happens when more than one hookimpl are marked as `tryfirst` or `trylast`? I believe then LIFO registered order for them applies, but perhaps this should be explicitly mentioned.

pluggy/callers.py
```diff
@@ -41,10 +41,21 @@ def from_call(cls, func):
         return cls(result, excinfo)
 
     def force_result(self, result):
+        """Force the result(s) to ``result``.
+
+        If the hook was marked as a ``firstresult`` a single value should
+        be set otherwise set a (modified) list of results. Any exceptions
+        found during invocation will be deleted.
+        """
         self.result = result
```
nicoddemus: Not related to the PR: should we make all attributes private?

tgoodlet: Hmm yeah maybe?
I don't think the original `_CallOutcome` did this any differently fwiw.

docs/index.rst
tgoodlet: Ok so something like:

> Note that `tryfirst` and `trylast` hooks are still invoked in LIFO order within each category

nicoddemus: Great 👍

pluggy/callers.py
nicoddemus: Oh definitely, this wasn't a criticism of the current code, just something I thought I might bring up. 

If we want to have a well-defined API we might want to start doing that somewhat sooner; a lot of problems of the pytest API stem from the fact that public attributes should never been so, but changing them later is a pain because client code now depends on it (even if by accident).

nicoddemus:
Excellent work btw, my comments are merely to improve this and that, overall everything looks great! Thanks for tackling this. 👍

pluggy/callers.py
tgoodlet: Yeah totally agree.
I just wasn't sure if `.result` was initially supposed to be public.
I'm cool to switch it in this PR as long as you guys are.

docs/index.rst
tgoodlet: done

tgoodlet:
oh @nicoddemus should I be flipping the version to `.dev` already here?

tgoodlet:
@nicoddemus good enough?

pluggy/callers.py
```diff
@@ -26,8 +26,12 @@ class HookCallError(Exception):
 
 class _Result(object):
     def __init__(self, result, excinfo):
-        self.result = result
-        self.excinfo = excinfo
+        self._result = result
```
nicoddemus: Just noticed that this broke [some tests](https://travis-ci.org/pytest-dev/pytest/jobs/274380302) in pytest, so pytest-dev/pytest#2744 is already paying off. 👍

PR #97, opened 2017-11-12
Prove that `PluginManger.register()` ignores non-routines

tgoodlet:
This proves that #91 is a problem in `pytest` and not `pluggy`.
Thanks to @RonnyPfannschmidt for the fix about a year ago ;)

### DIFF
--- a/testing/test_pluginmanager.py
+++ b/testing/test_pluginmanager.py
@@ -1,4 +1,5 @@
 import pytest
+import types
 
 from pluggy import (PluginValidationError,
                     HookCallError, HookimplMarker, HookspecMarker)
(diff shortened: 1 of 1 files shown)


testing/test_pluginmanager.py
```diff
@@ -350,3 +351,17 @@ def m(self, __multicall__, x):
...
         pm.add_hookspecs(10)
+
+
+def test_reject_prefixed_module(pm):
+    """Verify that a module type attribute that contains the project
+    prefix in its name (in this case `'example_*'` isn't collected
+    when registering a module which imports it.
+    """
+    pm._implprefix = 'example'
+    conftest = types.ModuleType("conftest")
```
nicoddemus: For completeness, I would add an `example_` function to one of the modules and ensure that it *is* collected.

tgoodlet: Done. Heh, and found a bug from the `_Multicall` refactor.. shocker we've got no tests for the undecorated hook collection stuff.

tgoodlet:
Hopefully the use of `exec()` doesn't fail on 2.6...

Issue #92, opened 2017-10-15
Drop py2.6 support


tgoodlet:
As the drop from `pytest` via pytest-dev/pytest#2812 we can now remove lingering code that is only in place to support old python.

See pytest-dev/pytest#2822 for comparable CI related changes that are likely appropriate. 

Some stuff I can think of off hand includes:
- set logic for [checking missing hookcall args](https://github.com/pytest-dev/pluggy/blob/master/pluggy/__init__.py#L672) can be modernized to use new `{,}` declarations and `dict.viewkeys()`.
- `str.format()` method now supports automatic numbering in py2.7+
- leveraging `collections.OrderedDict` to simplify some of the internal data structures for managing hooks

PR #103, opened 2017-11-20
Drop support for EOL Python 2.6 & 3.3

hugovk:
Fixes https://github.com/pytest-dev/pluggy/issues/92.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ matrix:
       env: TOXENV=check
     - python: '3.6'
       env: TOXENV=docs
-    - python: '2.6'
-      env: TOXENV=py26-pytestrelease
     - python: '2.7'
       env: TOXENV=py27-pytestrelease
     - python: '3.3'
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ matrix:
       env: TOXENV=docs
     - python: '2.7'
       env: TOXENV=py27-pytestrelease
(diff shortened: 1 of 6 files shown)


tgoodlet:
@hugovk looks good 👍 

Do you mind tagging your commit message with `Resolves #92` such that when we merge the issue will auto-close?

hugovk:
I certainly can do, but "Fixes #92" in the PR message should do the same:

https://help.github.com/articles/closing-issues-using-keywords/

Would you like the commit message updating?

nicoddemus:
@hugovk also please add a new CHANGELOG entry, I think the next version should be `0.6` because of this.

hugovk:
CHANGELOG updated.

Should we also drop EOL Python 3.3?

3.3 was dropped from pytest's features branch at the same time as 2.6: https://github.com/pytest-dev/pytest/pull/2822.

tgoodlet:
>I certainly can do, but "Fixes #92" in the PR message should do the same:

Cool, then I don't care too much if it's git history.

>3.3 was dropped from pytest's features branch at the same time as 2.6: pytest-dev/pytest#2822.

I'm 👍 on this but wait to see what @nicoddemus and @RonnyPfannschmidt say.

nicoddemus:
> Should we also drop EOL Python 3.3?

Yes please. 😁

hugovk:
Updated to drop 3.3 as well.

PR #108, opened 2017-11-24
Add CHANGELOG for #81

nicoddemus:


tgoodlet:
@nicoddemus shoot just saw this now.
Do you mind adding it to #109?

nicoddemus:
There's no need, you already mention this change in #109 here:

> - Fix ``inspect`` based arg introspection on py3.6 (PR `#94`_).

😉

PR #112, opened 2018-01-04
remove reference to missing 'excludefunc' arg in PluginManager docstring

anthrotype:
I noticed this while browsing the code. I don't see any `excludefunc` argument in PluginManager's constructor. Maybe it used to be there and was later removed?


nicoddemus:
Thanks!

PR #117, opened 2018-01-07
Split into modules

tgoodlet:
This is to replace #101 by @RonnyPfannschmidt.

The differences include:
- rebased onto master incorporating the latest changes
- maintaining as much of the history as possible in what was `pluggy.PluginManager` (where most of the non-@hpk42 contributions seemed to be) by moving `pluggy.__init__.py` -> `pluggy.manager` and pulling out internals from that module into other new modules
- ensuring the test suite can run successfully between each incremental change (commit) such that each patch is a logical change set
- avoiding the #101 proposed `parse_funcs` module (it's unecessary when it's a utility function used only by the code in `hooks.py`)


nicoddemus:
The split seems sensible to me... we just have to take a look at the test failures with pytest master.

tgoodlet:
>we just have to take a look at the test failures with pytest master

Yeah gonna try and reproduce locally with `tox`.
It's also weird that my latest for push hasn't shown up here in GH...

tgoodlet:
@nicoddemus hah! Think I figured it out.
The first time you run the `pytest` integration tests they fail because on install I think the old `pluggy` package is being used somehow. The failure is consistent as long as I pass `-r` to `tox`.

My original guess would be that the `_PYTEST_SETUP_SKIP_PLUGGY_DEP` isn't being set early enough since when I hop in the debugger in fact `pytest` is using the `pluggy` installed in the `virtualenv` instead of the sources. However, inspecting that env var shows it **is** set by the time it hits the import error?

Any ideas?

tgoodlet:
Did something change in `tox` regarding when/how the project under test is installed?

nicoddemus:
@tgoodlet a test of your theory would be to set `_PYTEST_SETUP_SKIP_PLUGGY_DEP` in the `.travis.yml` file manually before calling `tox`; if that changes the outcome then we know your hypothesis holds.

tgoodlet:
My [theory failed](https://travis-ci.org/pytest-dev/pluggy/jobs/326142225)...

I'm going to try and hook into my local `pytest` code and see what's going on.

tgoodlet:
Oh man @nicoddemus 🤦‍♂️ 
You're gonna laugh at me...

nicoddemus:
What is it? 😛

tgoodlet:
minor version bump was needed to get `pip` to install the sources...

nicoddemus:
Oh hehehe makes sense! Glad it was something simple then! 😁

tgoodlet:
Alright now I just need some approvals 👍

nicoddemus:
Nice work, thanks @tgoodlet!

tgoodlet:
@RonnyPfannschmidt @hpk42 you two ok with this?

PR #122, opened 2018-01-31
Add .plugin member to PluginValidationError to access failing plugin

ids1024:
The goal of this is to be able to call `.check_pending()` catching the exception, then unregister the plugin and print a message, allowing the program to continue without that plugin. If there is already a way to do this, I haven't found it.

### DIFF
--- a/pluggy/manager.py
+++ b/pluggy/manager.py
@@ -4,7 +4,10 @@ from .hooks import HookImpl, _HookRelay, _HookCaller, normalize_hookimpl_opts
 
 
 class PluginValidationError(Exception):
-    """ plugin failed validation. """
+    """ plugin failed validation. 
+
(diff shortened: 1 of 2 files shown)


pluggy/manager.py
```diff
@@ -6,6 +6,10 @@
 class PluginValidationError(Exception):
     """ plugin failed validation. """
```
nicoddemus: Please add this to the docs:

```python
""" plugin failed validation. 

:param object plugin: the plugin which failed validation, may be a module or an arbitrary object.
"""
```

nicoddemus:
Thanks!

tgoodlet:
Thanks @ids1024 can you please write a test for this which demonstrates and validates the feature.

PR #136, opened 2018-04-16
Improve docs: introduce roles and add another example

obestwalter:
My first go at explaining pluggy from the perspective of three different roles introducing all important concepts and practices in a reading time well under  5 minutes. Please note the pending fixme regarding dependency handling.


RonnyPfannschmidt:
IMHO pubsub is not a analogy of pluggy
while it is a hook system, callers control the receiving (path based force out in pytest for example) and the invocation
also modelling components like hook-wrappers, first-result and historic have a play

personally i believe a more fitting analogy is calling them **invocation of a call** and **participation in a call**
the dependency management could indeed take some help, its mainly a problem that python is fundamentally broken there

a lib like pluggy really could use support for having multiple versions installed in parallel - dito for things like attrs

RonnyPfannschmidt:
also thanks to things like first-result and hook-wrapper, there is more like 5-6 roles (and i am getting sadder and sadder as this unravels)

tgoodlet:
@obestwalter the description of *request-response pattern* and *publish-subscribe* was my doing in the initial docs draft. Agreed those descriptions aren't technically correct though I was trying to make a *kind of like* comparison:
>**thought of and used as** a **rudimentary busless** publish-subscribe_ **event system**.

Regarding,
>That is not an analogy. That is a rudimentary go at establishing some original terminology specific to pluggy. So you would suggest something like publisher -> invocator, subscriber -> participator? That sounds a bit esoteric to me but I have no better idea yet either. If we would go into that direction though, I would suggest to completely drop the uses of any analogies and their terminology and instead establish the pluggy concepts and terminology directly.

`invocator` or `caller` and `participator` or `implementer` works for me ; totally think we should go this way particularly now that it seems there's more users of the project outside of `pytest` and friends.

Please feel free to change/establish the terminology we'd like to use moving forward in this PR :+1:

docs/index.rst
```diff
@@ -1,8 +1,76 @@
...
+calls to their specified hooks in appropriate phases of program execution.
+``pluggy`` will then do all the magic.
+
+**FIXME not sure what is the right advice here. Do I need to vendor in pluggy as a publisher to avoid versioning conflicts? Should I just depend on pluggy without version and hope all is good?**
+To avoid compatibility issues they should always pin their ``pluggy`` version
+to the minor version used at implementation time (at the time of writing
+(spring 2018) ``pluggy`` is not at 1.0 yet, so breaking changes should be
+expected with any minor release).
+
+They need to learn about ``pluggy`` and its capabilities and should at least
```
tgoodlet: This snippet seems out of place?

docs/index.rst
```diff
@@ -1,8 +1,76 @@
...
+or modify some aspects of their programs execution. They write hook specifications,
+register them with a :py:class:`pluggy.PluginManager` and instantiate a marker
+object for the `subscriber` - a :py:class:`~pluggy.HookimplMarker` - used
+to `decorate <https://www.python.org/dev/peps/pep-0318/#current-syntax>` hook
+implementation functions. This marker is usually called `hookimpl` and
+importable from the root package of the `publisher` project. They then add
+calls to their specified hooks in appropriate phases of program execution.
+``pluggy`` will then do all the magic.
+
+**FIXME not sure what is the right advice here. Do I need to vendor in pluggy as a publisher to avoid versioning conflicts? Should I just depend on pluggy without version and hope all is good?**
```
tgoodlet: I'd say latter is what we recommend; so far I don't think we've had any breaking changes particularly because `pytest` and friends are all dependants.

docs/index.rst
```diff
@@ -1,8 +1,76 @@
...
+******************************
+
+``pluggy`` gives `users` the possibility to extend or modify the behaviour of
+a program like e.g. `pytest`_, `tox`_ or `devpi`_ by installing a `plugin` for
+that program. The plugin code will run as part of normal program execution
+changing or enhancing certain aspects of it.
+
+There are three roles in which people might interact with ``pluggy``.
+
+Publisher
```
tgoodlet: Maybe actually `Host` as in plugin host like described [here](https://en.wikipedia.org/wiki/Plug-in_%28computing%29#Mechanism)?

docs/index.rst
```diff
@@ -1,8 +1,76 @@
...
+**FIXME not sure what is the right advice here. Do I need to vendor in pluggy as a publisher to avoid versioning conflicts? Should I just depend on pluggy without version and hope all is good?**
+To avoid compatibility issues they should always pin their ``pluggy`` version
+to the minor version used at implementation time (at the time of writing
+(spring 2018) ``pluggy`` is not at 1.0 yet, so breaking changes should be
+expected with any minor release).
+
+They need to learn about ``pluggy`` and its capabilities and should at least
+loosely follow the development of the ``pluggy`` project.
+
+Subscriber
```
tgoodlet: Maybe `Provider` or `Implementer` as in plugin provider?

obestwalter:
Hey @tgoodlet - sounds good to me! I actually liked your analogies although they are unusual in the context we are in :)

So the actual explanation and the roles are o.k.? Anything missing. Formulation problems (I am not a native speaker and tend to use too long sentences, because german).

I will let this percolate for a while in the hope that maybe someone comes up with some better terminology and then run with whatever we have at that point.

docs/index.rst
```diff
@@ -1,8 +1,76 @@
...
+the `publisher` will discover the `plugin` hooks and start calling them as part
+of their normal program execution.
+
+If a `user` doesn't need the changed behaviour of a `plugin` anymore, they
+uninstall it and all is back to normal.
+
+`Users` don't need to know that the mechanism making this possible is provided by ``pluggy``.
+
+Origins: the ``pytest`` plugin system
+*************************************
```
tgoodlet: Feel free to change the downstream language as well :+1:

docs/index.rst
obestwalter: Its structured in a way that the last paragraph always gives info about the needed pluggy knowledge. I will make this explicit then.

docs/index.rst
obestwalter: o.k. that might be the sanest approach then :)

docs/index.rst
obestwalter: Yes! I caught myself writing host at several points, so this is the right way to go I guess. I will read up a bit more about plugin terminology also ...

docs/index.rst
obestwalter: implementer feels also right to me.

docs/index.rst
obestwalter: Thanks - that was basially just working around that I screwed up the structure by pluggin my intro on top. I will have a look at the whole thing then, when reworking the PR.

docs/index.rst
obestwalter: pytest does `'pluggy>=0.5,<0.7'` and devpi `pluggy>=0.3.0,<1.0` :scream:

RonnyPfannschmidt:
true, i approached this far too grumpy, sorry for unloading that on you, you certainly don't deserve that

i believe the terminology of "invoke" and "receive" are important for the elevator pitch as
an follow-up should eventually bring in details like first-result and hook-wrappers

participation plays more of a role when those concepts are introduced

obestwalter:
hmm ... invoke and receive don't seem to belong together for me. Also looking at roles that would be invoker and receiver - receiver is much to passive for me.

I am definitely with @tgoodlet on the `implementer` term, also because that term is already established and hinted at in the naming scheme (`hookimpl`, `HookImplMarker`) which the implementer then uses. So implementer is the way to go on that side IMO.

Leaves us with the other side. That seems to be the harder part, because if we would stick with internal names, we'd end up with a `specifier` and that just sounds odd - and to be clear: I am talking about naming roles for people here. A person called `specifier` sounds more like a biologist discovering new types of bugs and naming them. Discovering bugs is something we have in common with biologists, but we wouldn't want to be called specifier, would we? :D

As @tgoodlet said: there is already established terminlogy, so lets have a look in Wikipedia:

> The host application provides services which the plug-in can use, including a way for plug-ins to register themselves with the host application and a protocol for the exchange of data with plug-ins.

So the program is called `host` - that sounds right for the program, but I am not sure if calling the person working on that should be called `hoster`, but at least it's not terrible I guess, but maybe provider would be better?

> Programmers typically implement plug-in functionality using shared libraries, which get dynamically loaded at run time, installed in a place prescribed by the host application.

Programmers implement plugin functionality and a pogrammer doing that, is in the role of the implementer. Ease. See above.

So at the moment I lean towards hoster and implementer. I updated the PR to mirror that, so you can read it again and see how that rings.

docs/index.rst
```diff
@@ -3,17 +3,95 @@
...
 Some `200+ plugins`_ use ``pluggy`` to extend and customize ``pytest``'s default behaviour.
 
 In essence, ``pluggy`` enables function `hooking`_ so you can build "pluggable" systems.
 
-How's it work?
---------------
+Introduction
+------------
+
+To explain how pluggy can be put to use we need to introduce three roles and explain their interaction with pluggy:
```
tgoodlet: `To explain how ``pluggy`` is best utilized we need to introduce three use cases`:

A suggestion: reading this over again I think it might be more clear to talk about the code/programs themselves instead of people who write or use the code. This way we talk about what `host` versus `implementer` (maybe just `plugin`?) versus `user` (or `client`) code should look like and how it fits into the `pluggy` use-case model. 

Also, too many references to `pluggy` in a few places when we already know what's being referenced.

docs/index.rst
```diff
@@ -3,17 +3,95 @@
...
+
+* The `hoster` - the person who wants to add extensibility to their program
```
tgoodlet: The `host` - the program or project to be extended and augmented with plugins

docs/index.rst
```diff
@@ -3,17 +3,95 @@
...
+* The `implementer` - the person who wants to extend the `host` functionality by writing a plugin
```
tgoodlet: The `implementer` - the plugin itself which extends or modifies the `host`s functionality and/or behaviour

docs/index.rst
```diff
@@ -3,17 +3,95 @@
...
+* the `user` - the person who installed the `host` and now wants to extend its functionality by using a plugin
```
tgoodlet: the `user` - the program who depends on the `host` project and now wants to extend its functionality by using a plugin

docs/index.rst
```diff
@@ -3,17 +3,95 @@
...
+
+Hoster
+++++++
+
+The `hoster` wants to enable an arbitrary number of `implementers` to extend
```
tgoodlet: If we go with my suggestion the language will have to change a bit, eg:
`The host enables an arbitrary number of plugins to extend it's behaviour`

RonnyPfannschmidt:
true, we have implementation, specification, and invokation

obestwalter:
I think it's a good idea to set the roles like you suggested @tgoodlet - if @RonnyPfannschmidt is o.k. with the terms (host, implementer and user) for the fundamental roles, I will go ahead and adjust the docs that way.

RonnyPfannschmidt:
so far this is good work, i will try to review in detail this weekend, thanks!

RonnyPfannschmidt:
@obestwalter thanks for the note, please ping me later on when you feel confident in your reiteration

also :+1:

docs/index.rst
nicoddemus: In this regard I think we can all agree on pinning `>=0.x,<1.0`; we definitely don't want to introduce changes that break pytest, devpi or tox even in pre-1.0. 

And after all we are all the developers of the same projects, so we can all agree to postpone any breaking changes until 1.0.

docs/index.rst
nicoddemus: Good suggestions, I agree. 👍

docs/index.rst
RonnyPfannschmidt: i wouldn't want to put that limit on per se - but we certainly can coordinate fallout

technically we should probably do a 1.0 already since pluggy is in productive use, and ever since https://0ver.org/ its painfully oblivious that that is just wrong ^^

obestwalter: 0ver.org ... :laughing: reminds me of flask. We're using that in production since 2011 and it is at 0.12.2 atm.

nicoddemus: > technically we should probably do a 1.0 already since pluggy is in productive use,

Agree, let's make the next one `1.0` already then, or are there any significant features to be added before that?

obestwalter:
Hey @RonnyPfannschmidt and @tgoodlet - whaddaya think? It got a bit out of hand and I fixed a few more things along the way. Seems like small focussed PRs are not my strong suit ...

Also cc @fschulze: is that the right way to link to the devpi dev docs or is there a better way? I had to break out of the frame to get something that worked: https://github.com/pytest-dev/pluggy/pull/136/files#diff-caf2a6b8f4947d018f68893c695b5202R165

fschulze:
@obestwalter it's better to use "+d" instead of "+doc":
https://devpi.net/docs/devpi/devpi/stable/+d/devguide/index.html

There will be some documentation related fixes in the next devpi-web release regarding linking and fragments etc.

obestwalter:
Thanks @fschulze -fixed.

docs/examples/eggsample/eggsample/host.py
```diff
@@ -0,0 +1,32 @@
+import random
+
+from eggsample import hookimpl, hookspecs, lib, pm
+
+condiments_tray = {"pickled walnuts": 13, "steak sauce": 4, "mushy peas": 2}
```
tgoodlet: Man, gotta try me some pickled walnuts.

docs/examples/eggsample/eggsample/host.py
```diff
@@ -0,0 +1,32 @@
...
+    def serve_the_food(self):
+        pm.hook.eggsample_prep_condiments(condiments=condiments_tray)
+        print(f"Your food: {', '.join(self.ingredients)}")
+        print(f"Some condiments: {', '.join(condiments_tray.keys())}")
+
+def main():
+    pm.add_hookspecs(hookspecs)
+    pm.load_setuptools_entrypoints("eggsample")
+    pm.register(lib)
+    fm2000 = FoodMaster2000()
```
tgoodlet: Thought I love the name `FoodMaster2000` definitely seems like one of those cases where you ask *why is this a class and not just functions*?

docs/index.rst
```diff
@@ -1,63 +1,179 @@
...
+Why is it useful?
+*****************
+
+There are some established mechanisms for modifying the behavior of other
+programs/libraries in Python like
+`method overriding <https://en.wikipedia.org/wiki/Method_overriding>`_
+(e.g. Jinja2) or
+`monkey patching <https://en.wikipedia.org/wiki/Monkey_patch>`_ (e.g. gevent
+or for `writing tests <https://docs.pytest.org/en/latest/monkeypatch.html>`_).
+These strategies become problematic though when several parties want to
```
tgoodlet: Bravo on this description; well done!

docs/index.rst
```diff
@@ -1,63 +1,179 @@
...
+  participating in program execution when the implementations are invoked
+  by the ``host``
+* ``pluggy``: connects ``host`` and ``plugins`` by using ...
+
+    - the hook :ref:`specifications <specs>` defining call signatures
+      provided by the ``host`` (a.k.a ``hookspecs`` - see :ref:`marking_hooks`)
+    - the hook :ref:`implementations <impls>` provided by registered
+      ``plugins`` (a.k.a ``hookimpl`` - see `callbacks`_)
+    - the hook :ref:`caller <calling>` - a call loop triggered at appropriate
+      program positions in the ``host`` invoking the implementations
```
tgoodlet: `invoking the implementations **and** collecting the results`?

docs/index.rst
```diff
@@ -1,63 +1,179 @@
...
+Now let us demonstrate how this plays together in a vaguely real world scenario.
 
-- a hook *specification* - defines a call signature
-- a set of hook *implementations* - aka `callbacks`_
-- the hook *caller* - a call loop which collects results
+Let's assume our ``host program`` is called **eggsample** where some eggs will
+be prepared and served with a tray containing condiments. As everybody knows:
+the more cooks are involved the better the food, so let us make the process
+pluggable and write a plugin that improves the meal with some spam and removes
+the steak sauce from the condiments tray (nobody likes that anyway).
```
tgoodlet: Lolz definitely not on eggs.

docs/index.rst
```diff
@@ -1,63 +1,179 @@
...
 
-where for each registered hook *specification*, a hook *call* will invoke up to ``N``
-registered hook *implementations*.
+.. note::
 
-``pluggy`` accomplishes all this by implementing a `request-response pattern`_ using *function*
-subscriptions and can be thought of and used as a rudimentary busless `publish-subscribe`_
-event system.
+    **naming markers**: ``HookSpecMarker`` and ``HookImplMarker`` must be
+    the initialized with the name of the ``host`` project (the ``name``
```
tgoodlet: drop `the`

docs/index.rst
```diff
@@ -1,63 +1,179 @@
...
-Running this directly gets us::
+.. literalinclude:: examples/eggsample/eggsample/hookspecs.py
 
-    $ python docs/examples/firstexample.py
+``eggsample/eggsample/lib.py``
 
-    inside Plugin_2.myhook()
-    inside Plugin_1.myhook()
-    [-1, 3]
+.. literalinclude:: examples/eggsample/eggsample/lib.py
```
tgoodlet: I personally like including line numbers with [`:linenos:`](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=literalinclude#directive-literalinclude).

obestwalter:
I am getting quite eggsited now!

docs/examples/eggsample/eggsample/__init__.py
```diff
@@ -0,0 +1,7 @@
+import pluggy
+
+hookimpl = pluggy.HookimplMarker("eggsample")
+"""Marker to be imported and used in plugins"""
+
+pm = pluggy.PluginManager("eggsample")
```
RonnyPfannschmidt: should we take a note here that a pluginmanager instance is typically not a global ?

obestwalter: Yes, I was not happy with that bit either - I will see, if I can make it a bit more real worl like regarding this, so it is part of some mini config. If this adds too much cruft, I will at least add a comment.

docs/index.rst
obestwalter: Added it ... and removed it again. Looks absoulutely horrific due to https://github.com/sphinx-doc/sphinx/issues/2427 and does not really add too much utility IMO.

tgoodlet: Meh, not that big of a deal. Yeah just avoid the headache.

docs/examples/eggsample/eggsample/__init__.py
obestwalter: I added it in the main and pass it into the class similar to how its done in real world projects now.

docs/examples/eggsample/eggsample/host.py
obestwalter: Passed in pluginmanager and did it similar to how real world classes handle it in pytest or tox. Gives the class a reason to exist now (but I couldn't help but chaging the name using your pun :)).

tgoodlet:
@obestwalter where are we with this - all done?

@nicoddemus @RonnyPfannschmidt are you two happy?

obestwalter:
I'd say done ... except for maybe reading the complete docs as a whole making sure the terminology is consistent. I did not do that in this PR, as I think this one is huge enough. I would definitely like to have a go at this in another PR if you think it is a good idea.

obestwalter:
Hi @nicoddemus, thanks for having a look and get well soon.

docs/index.rst
```diff
@@ -1,63 +1,176 @@
...
+and objects to work with.
+
+How does it work?
+*****************
+Let us start with a short overview of what is involved:
+
+* ``host`` or ``host program``: the program offering extensibility
+  by specifying ``hook functions`` and invoking their implementation(s) as
+  part of program execution
+* ``plugin``: the program implementing a subset of the specified hooks and
```
tgoodlet: `a subset of` should maybe be inside `()` since technically a plugin could implement all hooks.
Though I guess the a set is a subset of itself?

obestwalter:
Hi @tgoodlet thanks for the comments - will fix that up over the weekend and hopefully it will be part of 1.0 then :)

docs/examples/eggsample-spam/eggsample_spam.py
```diff
@@ -0,0 +1,16 @@
...
+    if len([e for e in ingredients if e == "egg"]) > 2:
+        ingredients.remove("egg")
+    spam = ["lovely spam", "wonderous spam", "splendiferous spam"]
+    print(f"Add {spam}")
+    ingredients.extend(spam)
+
+@eggsample.hookimpl
+def eggsample_prep_condiments(condiments):
+    try:
+        del condiments["steak sauce"]
```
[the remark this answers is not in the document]
obestwalter: You think so? I'd say you use `del` if you want to get rid of something and `list.pop` if you want to do something with the value. Here the steak sauce is just removed from the tray to vanish into thin air, so I'd use `del` here.

docs/examples/eggsample/eggsample/host.py
```diff
@@ -0,0 +1,36 @@
+import random
+
+import pluggy
+
+from eggsample import hookspecs, lib
+
+condiments_tray = {"pickled walnuts": 13, "steak sauce": 4, "mushy peas": 2}
+
+class EggselentCook:
```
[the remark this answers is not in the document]
obestwalter: oh yeah, thanks :)

docs/examples/eggsample/eggsample/host.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, hook):
+        self.hook = hook
+        self.ingredients = []
+
+    def add_ingredients(self):
+        self.hook.eggsample_add_ingredients(ingredients=self.ingredients)
```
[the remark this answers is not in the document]
obestwalter: Good idea, will adjust this.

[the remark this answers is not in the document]
obestwalter: My original intention here was to make clear that due to Pythons behaviour of passing by assignment it is possible to mutate the object that pluggy passes to the client, which is also used regularly e.g. when adding command line parameters to the parser, but I agree that in the context of pluggy it is more important to showcase the return behaviour.

docs/index.rst
```diff
@@ -1,63 +1,176 @@
...
+These strategies become problematic though when several parties want to
+participate in the modification of the same program. Therefore ``pluggy``
+does not rely on these mechanisms to enable a more structured approach and
+avoid unnecessary exposure of state and behaviour. This leads to a more
+`loosely coupled <https://en.wikipedia.org/wiki/Loose_coupling>`_ relationship
+between ``host`` and ``plugins``.
+
+The ``pluggy`` approach puts the burden on the designer of the
+``host program`` to think carefully about which objects are explicitly needed
+by an implementation and gives designers of the ``plugin`` a clear framework
```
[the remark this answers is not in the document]
obestwalter: Much better. Thanks.

docs/index.rst
obestwalter: I had the same thoughts myself and came to the latter conclusion, but if you had to do the same mental gymnastics, I guess it is better to add the parens :)

docs/index.rst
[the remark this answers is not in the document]
obestwalter: I tweaked it a bit more, hope it's ok like this.

obestwalter:
@tgoodlet - thanks for the suggestions, from my side this is good to go now.

obestwalter:
I can be awfully repetitive sometimes - but I guess you can never say thanks too often :)

obestwalter:
o.k. unless I managed to improve another bug, while improving it, I would say this is good to go.

tgoodlet:
@RonnyPfannschmidt do you need one more look over or can we merge this?

nicoddemus:
Definitely, awesome work @obestwalter and @tgoodlet on reviewing! 👍

PR #143, opened 2018-05-17
Deprecate `proc` kwarg to call_historic()

### LINKED ISSUE #120: Change call_historic callback kwarg name
As per discussion in #110 we agreed that the kwarg `proc` in `_HookCaller.call_historic()` is poorly named and should be instead renamed `result_callback`. 

`proc` should be deprecated in a minor release with an appropriate warning and later removed.

tgoodlet:
It's a bad name, use `result_callback` instead.

Resolves #120

### DIFF
--- a/pluggy/hooks.py
+++ b/pluggy/hooks.py
@@ -257,20 +257,32 @@ class _HookCaller(object):
                 )
         return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
 
-    def call_historic(self, proc=None, kwargs=None):
-        """ call the hook with given ``kwargs`` for all registered plugins and
+    def call_historic(self, result_callback=None, kwargs=None, proc=None):
+        """Call the hook with given ``kwargs`` for all registered plugins and
         for all plugins which will be registered afterwards.
 
-        If ``proc`` is not None it will be called for for each non-None result
-        obtained from a hook implementation.
+        If ``result_callback`` is not ``None`` it will be called for for each
+        non-None result obtained from a hook implementation.
+
+        .. note::
+            The ``proc`` argument is now deprecated.
         """
-        self._call_history.append((kwargs or {}, proc))
+        if proc is not None:
+            warnings.warn(
+                "Support for `proc` argument is now deprecated and will be"
+                "removed in an upcoming release.",
+                DeprecationWarning
+            )
+            result_callback = proc
+
+        self._call_history.append((kwargs or {}, result_callback))
         # historizing hooks don't return results
         res = self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
-        if proc is None:
+        if result_callback is None:
             return
+        # XXX: remember firstresult isn't compat with historic
         for x in res or []:
-            proc(x)
+            result_callback(x)
 
     def call_extra(self, methods, kwargs):
         """ Call the hook with some additional temporarily participating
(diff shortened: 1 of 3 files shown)


pluggy/hooks.py
```diff
@@ -257,20 +257,29 @@ def __call__(self, *args, **kwargs):
...
-    def call_historic(self, proc=None, kwargs=None):
-        """ call the hook with given ``kwargs`` for all registered plugins and
+    def call_historic(self, result_callback=None, kwargs=None, proc=None):
+        """Call the hook with given ``kwargs`` for all registered plugins and
         for all plugins which will be registered afterwards.
 
-        If ``proc`` is not None it will be called for for each non-None result
-        obtained from a hook implementation.
+        If ``result_callback`` is not ``None`` it will be called for for each
+        non-None result obtained from a hook implementation.
```
nicoddemus: Please document that `proc` is deprecated in the docstring as well.

testing/test_pluginmanager.py
```diff
@@ -217,7 +217,13 @@ def he_method1(self, arg):
     pm.register(Plugin1())
 
     he_method1 = pm.hook.he_method1
-    he_method1.call_historic(proc=callback, kwargs=dict(arg=1))
+
+    # TODO: remove once ``proc`` is removed
+    if callback:
```
nicoddemus: For completeness we should also test passing `result_callback`. I suggest changing this test to use `result_callback` and create a new, simpler test, just for the deprecation.

pluggy/hooks.py
tgoodlet: Will do!

testing/test_pluginmanager.py
tgoodlet: Sure thing!

tgoodlet:
@nicoddemus pushed all your recs.

pluggy/hooks.py
```diff
@@ -257,20 +257,32 @@ def __call__(self, *args, **kwargs):
...
+            )
+            result_callback = proc
+
+        self._call_history.append((kwargs or {}, result_callback))
         # historizing hooks don't return results
         res = self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
-        if proc is None:
+        if result_callback is None:
             return
+        # XXX: woah isn't this wrong for `firstresult` hooks?
```
tgoodlet: @nicoddemus btw did you notice this?
If i'm reading this correctly won't this code never process `firstresult` hooks?

tgoodlet: Yeah I think that loop expr is just slightly wrong should be `for x in res or [res]:` right?

pluggy/hooks.py
```diff
@@ -257,20 +257,32 @@ def __call__(self, *args, **kwargs):
...
+        If ``result_callback`` is not ``None`` it will be called for for each
+        non-None result obtained from a hook implementation.
+
+        .. note::
+            The ``proc`` argument is now deprecated.
         """
-        self._call_history.append((kwargs or {}, proc))
+        if proc is not None:
+            warnings.warn(
+                "Support for `proc` argument now deprecated and will be"
```
tgoodlet: `Support for `proc` **is** argument now`  dahhh!

pluggy/hooks.py
tgoodlet: Oh waitttt [whatttt](https://github.com/pytest-dev/pluggy/blob/master/pluggy/hooks.py#L35)?
Wait why not?

tgoodlet:
@nicoddemus @RonnyPfannschmidt give this one more lookover if you don't mind. I tweaked couple things and added docs.

Issue #100, opened 2017-11-13
Test modules named incorrectly?


tgoodlet:
I've noticed that [test_method_ordering.py](https://github.com/pytest-dev/pluggy/blob/master/testing/test_method_ordering.py) and [test_hookrelay.py](https://github.com/pytest-dev/pluggy/blob/master/testing/test_method_ordering.py) might want to have their names switched based on the test contents of each.

tgoodlet:
Looking at this closer I think a better name for `test_hookrelay.py` is `test_invocations.py`.

tgoodlet:
Looking at this closer I think a better name for `test_hookrelay.py` is `test_invocations.py` and `test_method_ordering.py` is better as `test_hookcaller.py`.

Some tests need to be moved around as well.

PR #153, opened 2018-05-22
Tests reorg

tgoodlet:
Addresses #100 as well as attempts to bring a little more consistency and appropriate organization to our test module's content. See the commit msg in 4c19d8d for the details on the two module (re)namings.
This should at the least make it easier to find tests related to specific components and help us avoid duplication (of which I still think there is some but fixing that will come later).

I also moved a bunch of tests from the `test_method_ordering` (now `test_hookcaller`) into `test_pluginmanager` as they really are just tests of the public API for both tracing and the `setuptools` entry point loading stuff.

I added a specific module for deprecation warnings tests.

There's further *deprecation* tests in `test_pluginmanager.py` now that #144 has landed.
I wasn't sure if it'd be appropriate to move these into `test_deprecations` now or if that module should be strictly for warnings tests.

`test_details` is also something we might want to rename as it doesn't convey much meaning.


tgoodlet:
@RonnyPfannschmidt @nicoddemus this is ready for viewing I believe.

PR #160, opened 2018-07-27
Release 0.7.1

nicoddemus:
OK now we are in business.

Had to call `git commit` directly in the commit script because apparently gitpython bypasses the local configuration (don't have time right now to look for another solution).

Other than that, I changed #131 and #66 from `feature` to `trivial`, as those are more internal changes than user-facing features.

PR #164, opened 2018-07-27
Try to reproduce the problem in #160

nicoddemus:


nicoddemus:
Oh OK, after fixing the exception handling, it explains the error in #160 then:

```
pluggy.manager.PluginValidationError: Plugin 'benchmark' could not be loaded: (pluggy 0.7.0 (/home/travis/build/pytest-dev/pluggy/.tox/benchmark/lib/python3.6/site-packages), Requirement.parse('pluggy<0.7,>=0.5'), {'pytest'})!
```

nicoddemus:
Hmm actually no. `pytest-benchmark`, unsurprisingly, does not pin `pluggy` at all.

I'm guessing it is a problem with `pip` package resolution.

nicoddemus:
Cherry-pick the fix in #160.

PR #167, opened 2018-07-30
Normalize all titles and add dates in CHANGELOG

nicoddemus:


nicoddemus:
Twice approved even 😜 

Merging then! 😁

PR #172, opened 2018-08-08
Add a test demonstrating the opt in nature of args

tgoodlet:
Args are always opt in for hookimpls regardless of whether the spec or
call defines more then the impl accepts.

Relates to #170


tgoodlet:
@nicoddemus so changelog entry for this is just a `.trivial` yah?

PR #181, opened 2018-10-16
Adjust stacklevel of implprefix warning

### LINKED ISSUE #166: Adjust the implprefix DeprecationWarning's stack level
Hi!

In #144 a deprecation warning was added for usage of the `implprefix` argument.

However the `warnings.warn()` call doesn't set a `stacklevel`, which means when the warning is output, the blamed line is in pluggy, rather than the plugin that's using `implprefix`. For example:
https://travis-ci.org/mozilla/treeherder/jobs/409417404#L895

Passing `stacklevel=2` would make it clearer as to where the issue lies, and also mean the correct module path can be referenced in eg pytest's `filterwarnings` (for downstream projects who enable warnings/deprecation warnings as errors).

Many thanks :-)

CC @tgoodlet

crazymerlyn:
Fixes #166.

### DIFF
--- a/changelog/166.trivial.rst
+++ b/changelog/166.trivial.rst
@@ -1,0 +1,1 @@
+Add ``stacklevel=2`` to implprefix warning so that the reported location of warning is the caller of PluginManager.
--- a/pluggy/manager.py
+++ b/pluggy/manager.py
@@ -53,6 +53,7 @@ class PluginManager(object):
                 "Support for the `implprefix` arg is now deprecated and will "
                 "be removed in an upcoming release. Please use HookimplMarker.",
                 DeprecationWarning,
+                stacklevel=2,
             )
         self._implprefix = implprefix
         self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
--- a/testing/test_pluginmanager.py
+++ b/testing/test_pluginmanager.py
@@ -579,6 +579,12 @@ def test_hook_tracing(he_pm):
         undo()
 
 
+def test_implprefix_warning(recwarn):
+    PluginManager(hookspec.project_name, "hello_")
+    w = recwarn.pop(DeprecationWarning)
+    assert "test_pluginmanager.py" in w.filename
+
+
 @pytest.mark.parametrize("include_hookspec", [True, False])
 def test_prefix_hookimpl(include_hookspec):
     with pytest.deprecated_call():


crazymerlyn:
The build is failing due to a recent [change](https://github.com/pytest-dev/pytest/pull/4104/commits/e0f6fce9e9213d5beda7f54e800a72ac95cbb3b6#diff-f08479569882d0799350ec99c3b8a05eR158) in pytest which adds a wrapper for `warnings.warn` function for python 2 (thus, changing the stack). Not sure what the best solution is here.
Adding a simple `if 'stacklevel' in kwargs: kwargs['stacklevel'] += 1` in the wrapper function could work.

tgoodlet:
@crazymerlyn thanks for the PR :+1:!
@nicoddemus might know the right tweak to make.

nicoddemus:
> Adding a simple if 'stacklevel' in kwargs: kwargs['stacklevel'] += 1 in the wrapper function could work.

Oh indeed, good catch. What do you think @asottile?

asottile:
> > Adding a simple if 'stacklevel' in kwargs: kwargs['stacklevel'] += 1 in the wrapper function could work.
> 
> Oh indeed, good catch. What do you think @asottile?

oh! I see what you're saying now -- I was wracking my brain to figure out what was meant here.

```python
if isinstance(kwargs.get('stacklevel'), int):
    kwargs['stacklevel'] += 1
```

might be safer

probably PR that to pytest and we can get it into 3.9.2?

crazymerlyn:
> ```python
> if isinstance(kwargs.get('stacklevel'), int):
>     kwargs['stacklevel'] += 1
> ```
> 
> might be safer
> 
> probably PR that to pytest and we can get it into 3.9.2?

Yeah, that seems better. But while we are on this, I think stacklevel also needs to be changed if it is not present in kwargs. The default value is 1. So, if somebody uses recwarn in python2.7, the warning will present itself at the wrapper function rather than the line causing the warning.
How about this:
```python
if isinstance(kwargs.get('stacklevel'), int):
    kwargs['stacklevel'] += 1
elif 'stacklevel' not in kwargs:
    kwargs['stacklevel'] = 2
```

asottile:
ah in that case even easier:

```python
kwargs.setdefault('stacklevel', 1)
kwargs['stacklevel'] += 1
```

asottile:
hmm yeah this is probably the `__warningregistry__` bits.  link your branch? I'll poke at it since I broke this :sob:

nicoddemus:
> I'll poke at it since I broke this 

Heh, that's all good and part of the game 😁

crazymerlyn:
@asottile https://github.com/crazymerlyn/pytest/tree/fix_recwarn_stacklevel.

asottile:
not pretty, but at least this is throwaway :'( https://github.com/pytest-dev/pytest/pull/4192

asottile:
sorry about the trouble again, thanks for the report! 🎉

tgoodlet:
@nicoddemus good to go here yah?

asottile:
yah!

edmorley:
Many thanks!

PR #184, opened 2018-11-09
Release 0.8.1

nicoddemus:


tgoodlet:
Heh, totally missed this. Anything else we want to get in first?

nicoddemus:
Oh me too. 

IMHO we don't need to worry too much about which fixes go in patch releases, we can always make a new patch release later. Minor releases require a little more planning though.

PR #195, opened 2019-03-01
Fix method name `add_hookspecs` in docstring

jakobkogler:
Correct the method name `add_hookspec` to `add_hookspecs` in the docstring of the `PluginManager` class.


nicoddemus:
Thanks!

Issue #140, opened 2018-04-25
Release 1.0?


nicoddemus:
Hi,

As has been discussed before, is it time to start discussing about the 1.0 release? It has been in use in production in `pytest`, `tox` and `devpi` for some time after all.

* Do we have any major changes that are backward incompatible?
* What to do about our scheduled [deprecations](https://github.com/pytest-dev/pluggy/labels/deprecation)?

### Milestone 1.0
https://github.com/pytest-dev/pluggy/milestone/1

tgoodlet:
@nicoddemus sorry must have missed this. Yes I guess we should start working seriously towards this.

I was kind of hoping for at least the following features to be brought in before `1.0`:
- #51
- #87

This would lay a clean-ish slate for moving onto bigger and better ideas.
It's my bad for not being more on top of this; I've been somewhat distracted with other projects recently.

I would love to see the deprecation list finished up as well yes!
I'd be interested in setting some milestones so that we can start pushing towards this.

RonnyPfannschmidt:
#51 is too big and complicated to get right in any kind of rush
#87 is "just a feature" and needs experimentation in any case to get the api level right (see the discussion there)

we need to reiterate introducing/removing arguments for invocations over time - i'd like to have @hpk42 as part of that discussion

hpk42:
I haven't tracked pluggy development lately. 
FWIW a 1.0 sounds fine with me, it's indeed stable 
enough and there can always be a 1.1 and 2.0.

nicoddemus:
I agree with @hpk42, IMO we only need to hold a 1.0 release if we have plans for backward incompatible changes for the near feature or if there are things we should start deprecating now, otherwise new features can come in the next minor releases.

The main thing for me is that we have been using `pluggy` in production for years, so it is definitely production-ready.

RonnyPfannschmidt:
i agree - thanks everyone for the input

tgoodlet:
So should we move forward with the deprecation list then now or no?
I think it'd be good to get some of the cruft flushed out of the code base if at all possible - less is more some say :)

nicoddemus:
By deprecation list you mean https://github.com/pytest-dev/pluggy/labels/deprecation?

* #116 Deprecate (and remove) implprefix support - pytest can't really live without this, so as we discussed we should implement it into `PytestPluginManager` and remove it from pluggy.
[...]

PR #199, opened 2019-04-07
Switch to importlib-metadata

asottile:
This attempts to switch the backing metadata looking from `pkg_resources` (slow) to `importlib-metadata` (fast)

I had to expose one derived property on top of the `importlib_metadata` `Distribution` class to make it look like the `pkg_resources` one -- but otherwise this should be drop-in.

At least for pytest this isn't enough to get the full performance boost since `pytest` imports `pkg_resources` itself -- but baby steps!  I'll be following up with the pytest branch in https://github.com/pytest-dev/pytest/pull/5063

### DIFF
--- a/testing/test_pluginmanager.py
+++ b/testing/test_pluginmanager.py
@@ -471,7 +471,12 @@ def test_load_setuptools_instantiation(monkeypatch, pm):
     assert num == 1
     plugin = pm.get_plugin("myname")
     assert plugin.x == 42
-    assert pm.list_plugin_distinfo() == [(plugin, dist)]
+    ret = pm.list_plugin_distinfo()
+    # poor man's `assert ret == [(plugin, mock.ANY)]`
+    assert len(ret) == 1
+    assert len(ret[0]) == 2
+    assert ret[0][0] == plugin
+    assert ret[0][1]._dist == dist
     num = pm.load_setuptools_entrypoints("hello")
     assert num == 0  # no plugin loaded by this call
 
--- a/testing/test_pluginmanager.py
+++ b/testing/test_pluginmanager.py
@@ -3,7 +3,7 @@
 """
 import pytest
 import types
-import sys
+import importlib_metadata
 from pluggy import (
     PluginManager,
     PluginValidationError,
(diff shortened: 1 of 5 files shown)


asottile:
This is WIP:

- [ ] the `dist` objects that are exposed by the pluggy api currently are `pkg_resources` objects
- [ ] pytest uses `dist.project_name` which isn't a thing with `importlib-metadata` `Distribution` objects

need to make a shim?  `dist.metadata['name']` ~= `dist.project_name`

asottile:
This is no longer WIP, technically as far as I can tell this version is compatible with pytest as-is (due to the `Distribution` shim)

asottile:
switched the last commit from `python -m tox` to `tox` directly, it looks like until `pluggy` is the `src` layout tox is importing it directly 🤦‍♂️

testing/test_pluginmanager.py
```diff
@@ -447,64 +447,40 @@ def example_hook():
...
+    # poor man's `assert ret == [(plugin, mock.ANY)]`
+    assert len(ret) == 1
+    assert len(ret[0]) == 2
+    assert ret[0][0] == plugin
+    assert ret[0][1]._dist == dist
     num = pm.load_setuptools_entrypoints("hello")
     assert num == 0  # no plugin loaded by this call
 
 
-def test_load_setuptools_version_conflict(monkeypatch, pm):
```
jayvdb: runtime checking of dependency versions is not possible without the old and slow pkg_resources? ;-(

asottile: hmmm I guess that's not the way I saw that.  any time I've dealt with `pkg_resources` the side-effect of looking up package metadata triggering `VersionConflict` I always saw as a nuisance.  It's not really protecting you against anything, you're already in the broken state.

It would be possible to do with importlib-metadata, but traversing the dependency graph is wasted cpu cycles :)

nicoddemus:
Anything else to be done here @asottile? Also, can we check how this affects `tox` and `devpy`?

cc @gaborbernat @fschulze

asottile:
I think testing the integration with the other big users is all that's left here.  my pytest PR passes tests when coupled with this PR (unfortunately pytest has some tests which test the full end-to-end integration and worry about some specifics of the implementation of pluggy and so changes were necessary there)

asottile:
tox's testsuite passes, there's not much that tox does directly with `pkg_resources` and would be easy to factor that out 👍

fschulze:
Does it affect users of the pluggy library in any way? By the looks of it, it shouldn't. And the things we use from pkg_resources in devpi should be covered by importlib-metadata as well, so maybe we will also switch. I think only in tests are we using some things that aren't available yet.

asottile:
there _shouldn't_ affect users in any way (other than potentially speeding things up if `pkg_resources` gets factored out of the rest of the things) -- the only changes I had to make for pytest to make the tests pass was adjusting the tests (though I did also factor out most of the rest of the pytest usages of pkg_resources to get the full speed boost)

nicoddemus:
Great!

I propose we merge this and make a `0.10` release then. We can then update pytest's PR and get this out with pytest 4.5. What do you think @asottile?

asottile:
Sounds good! Let's do this Tuesday at the earliest (busy with pycon today and Monday and would want to be a available for potential fixes)

gaborbernat:
tox will follow probably with a release on Thursday, to take advantage of the pytest release lessons 😎

nicoddemus:
> Let's do this Tuesday at the earliest (busy with pycon today and Monday and would want to be a available for potential fixes)

You mean pluggy, pytest or both? My original intention was to release `pluggy` soon only. pytest still has a few open PRs I would like to see merged before we release 4.5.

nicoddemus:
(btw jealous of you guys at PyCon ❤️ 😁)

asottile:
> > Let's do this Tuesday at the earliest (busy with pycon today and Monday and would want to be a available for potential fixes)
> 
> You mean pluggy, pytest or both? My original intention was to release `pluggy` soon only. pytest still has a few open PRs I would like to see merged before we release 4.5.

I was mostly referring to pluggy (and with the hope that pytest would follow soon after, I can help finishing up the other things we want to get into 4.5 as well :D)

nicoddemus:
OK, let's wait until Tuesday then! 👍

nicoddemus:
Oops forgot about the changelog entry.

fschulze:
What about making it a 1.0.0? It's been in use for quite a while now.

goodboy:
@fschulze we're trying to get in a bunch of "breaking" changes for `1.0` which I can hopefully get to at some point in the next month. Unfortunately been really busy with non-software stuff atm..

nicoddemus:
We are tracking `1.0` here: https://github.com/pytest-dev/pluggy/issues/140

asottile:
I've re-applied this in #215 🎉

gaborbernat:
cool times

cool times

PR #212, opened 2019-05-14
Preparing release 0.11.1

nicoddemus:


hroncok:
Before this is released, plese see https://github.com/pytest-dev/pluggy/pull/210#issuecomment-492712324

nicoddemus:
Rebased. 👍

goodboy:
@nicoddemus we've got that [benchmark suite](https://github.com/pytest-dev/pluggy/blob/master/tox.ini#L18) that could be used to eyeball for slowdowns?

nicoddemus:
Here are the benchmarks, what do you guys think?

https://gist.github.com/nicoddemus/105e7bd20b9699ac5330968fadfc5b67

Also just realized this should be `0.12.0` because of the new dependency.

goodboy:
@nicoddemus slightly faster on `master`?

Wait, `0.11.1` is the release you just made - so we got a slight slow down I guess?
If that's the case then this is what we expect yes?

nicoddemus:
`0.11.1` is this branch (not yet released), so yeah we got a little slower. Do people think this is a blocker and we should revert the patch? IMHO no, but would like to hear others' opinions.

nicoddemus:
Gentle ping @goodboy and @asottile 😁

goodboy:
@nicoddemus I think it's probably fine; we're still getting most of our speedups by dropping the old multicall stuff.

If they reverted that deprecated in `cpython` I wonder if it's worth all this effort though.

hroncok:
> If they reverted that deprecation in cpython I wonder if it's worth all this effort though.

Good question. Does the new code make it more readable or maintainable?

asottile:
With the undeprecation I'm not sure moving to Signature is buying us much -- I don't particularly care either way though so: +0

nicoddemus:
> Good question. Does the new code make it more readable or maintainable?

TBH it is a little more complicated, but only because we still need to maintain two APIs: `inspect.getsignature` and `inspect.Signature`, which are quite different. It is also slightly slower.

Given that the deprecation is removed, I think it would be a good idea to revert the patch (if @hroncok doesn't mind, of course).

hroncok:
My only motivation was to get rid of the warning. Of course I don't mind.

But please keep the added test.

nicoddemus:
Thanks @hroncok! Opened https://github.com/pytest-dev/pluggy/pull/214.

PR #216, opened 2019-05-27
Preparing release 0.12.0

asottile:


nicoddemus:
Feel free to announce this on Twitter if you like! 🎆

nicoddemus:
You might mention that pytest 4.5 users will get speedup this automatically (https://github.com/pytest-dev/pytest/blob/master/setup.py#L17)

asottile:
pytest 4.5 users won't get the speedup just yet as pytest uses `pkg_resources` itself -- though I did make sure that this version of pluggy still works with current versions of pytest :)

nicoddemus:
Oh duh right. 👍

Issue #222, opened 2019-07-02
Use standard library importlib.metadata on Python 3.8


hroncok:
Once we update Python in Fedora to 3.8, we'd like our pluggy package to use `importlib.metadata` from the standard library. Would you accept a PR that makes the dependency on external `importlib_metadata` optional (only use it on Python < 3.8)?

nicoddemus:
I defer the final decision to @asottile, but it seems reasonable to me! 👍

asottile:
👍

matejcik:
Please release a fix for this issue. This now breaks pluggy and pytest on Python 3.8 with `importlib_metadata` 0.21 -- see https://gitlab.com/python-devs/importlib_metadata/issues/86

nicoddemus:
0.13 should be released soon: https://github.com/pytest-dev/pluggy/pull/234

PR #223, opened 2019-07-02
Replace importlib_metadata with importlib.metadata on Python 3.8+

hroncok:
Fixes https://github.com/pytest-dev/pluggy/issues/222

Got `pytest: error: unrecognized arguments: -ra` with tox on 3.8 :(

### DIFF
--- a/testing/test_pluginmanager.py
+++ b/testing/test_pluginmanager.py
@@ -2,8 +2,9 @@
 ``PluginManager`` unit and public API testing.
 """
 import pytest
+import sys
 import types
-import importlib_metadata
+
 from pluggy import (
     PluginManager,
     PluginValidationError,
@@ -11,6 +12,11 @@ from pluggy import (
     HookimplMarker,
     HookspecMarker,
 )
+
+if sys.version_info >= (3, 8):
+    from importlib import metadata
+else:
+    import importlib_metadata as metadata
 
 
 hookspec = HookspecMarker("example")
@@ -466,7 +472,7 @@ def test_load_setuptools_instantiation(monkeypatch, pm):
     def my_distributions():
         return (dist,)
 
-    monkeypatch.setattr(importlib_metadata, "distributions", my_distributions)
+    monkeypatch.setattr(metadata, "distributions", my_distributions)
     num = pm.load_setuptools_entrypoints("hello")
     assert num == 1
     plugin = pm.get_plugin("myname")
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,16 +32,16 @@ jobs:
       env: TOXENV=py37-pytestrelease-coverage
     - python: '3.8-dev'
       env: TOXENV=py38-pytestrelease-coverage
-    - python: '2.7'
-      env: TOXENV=py27-pytestmaster-coverage
-    - python: '2.7'
-      env: TOXENV=py27-pytestfeatures-coverage
     - python: '3.6'
       env: TOXENV=py36-pytestmaster-coverage
     - python: '3.6'
       env: TOXENV=py36-pytestfeatures-coverage
     - python: '3.6'
       env: TOXENV=benchmark
+    - python: '3.7'
+      env: TOXENV=py37-pytestmaster-coverage
+    - python: '3.7'
+      env: TOXENV=py37-pytestfeatures-coverage
 
     - stage: deploy
       python: '3.6'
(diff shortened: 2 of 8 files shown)


nicoddemus:
> Got pytest: error: unrecognized arguments: -ra with tox on 3.8 :(

Yes this is a known bug in 3.8, which was fixed by https://github.com/pytest-dev/pytest/issues/5523 but not yet released. Could you please change `-ra` to `-r a`?

testing/test_pluginmanager.py
```diff
@@ -3,7 +3,12 @@
 """
 import pytest
 import types
-import importlib_metadata
+
+try:
```
nicoddemus: Would you mind changing this to:

```python
if sys.version_info >= (3, 8):
    from importlib import metadata
else:
    import importlib_metadata as metadata
```

?

This makes it explicit which version this is supposed to work (so we can eventually find and remove the obsolete code in the future when <py37 support is dropped), and also avoids hiding unrelated bugs in broken environments which could cause `from importlib import metadata` to raise an `ImportError` even on Python 3.8.

nicoddemus:
> Got pytest: error: unrecognized arguments: -ra with tox on 3.8 :(

Yes this is a known bug in 3.8, which was fixed by https://github.com/pytest-dev/pytest/issues/5523 but not yet released. Could you please change `-ra` to `-r a`?

nicoddemus:
Thanks @hroncok! 👍

hroncok:
OK. The tox/travis configuration also probably needs some teaks:

 * there is no 3.8 on travis
 * there is no 3.7 nor 3.8 in tox
 * py27-pytest{master,features} no longer works

testing/test_pluginmanager.py
hroncok: I could, however I always thought that feature detection is preferred.

nicoddemus:
Agreed, we should update the configuration after pytest 5.0 has been released.

You want to do the tweaks yourself, or would you like me to do it?

hroncok:
Do you mind me pushing it here?

blueyed:
> there is no 3.8 on travis

It is?!

Apart from that makes sense to add it here.

testing/test_pluginmanager.py
nicoddemus: I prefer the explicit version because of the aforementioned reasons. It was a breeze to remove Python 2.7 only features when they were explicitly marked using `six.PY2` or `sys.version_info`, but the ones using feature detection were much harder (and I doubt we got them all, probably there are others in place still).

hroncok:
> there is no 3.8 on travis

yep, crossed that out right after I've posted it. sorry for the confusion

nicoddemus:
> Do you mind me pushing it here?

Not at all, but if possible make that in a separate commit. 👍

testing/test_pluginmanager.py
hroncok: What if I add a comment there? (Just to the actual source, not to tests and docs).

As for broken environments, that would be broken anyway, as importlib_metadata would not be installed.

hroncok:
Not sure if appveyor has 3.8, we'll see.

testing/test_pluginmanager.py
hroncok: however, I can definitively change that, yes.

nicoddemus: > What if I add a comment there? (Just to the actual source, not to tests and docs).

If you don't mind I would prefer the explicit version check, sorry for being insistent on that. 😁 

> As for broken environments, that would be broken anyway, as importlib_metadata would not be installed.

Yes, but then we will get a message like "importlib not available because <broken environment>" (the real problem), instead of "Could not import importlib_metadata", which would be hide the original error and would be baffling (experienced this myself more than once unfortunately).

hroncok: amended

hroncok:
appveyor: ERROR: InterpreterNotFound: python3.8

nicoddemus:
Thanks @hroncok!

Do you need a new release with this change?

hroncok:
Thanks for help. No if it is not needed for https://github.com/pytest-dev/pytest/issues/5537

testing/test_pluginmanager.py
```diff
@@ -12,6 +13,11 @@
     HookspecMarker,
 )
 
+if sys.version_info >= (3, 8):
+    from importlib import metadata
+else:
+    import importlib_metadata as metadata
+
```
blueyed: Just wondered if this could/should import/use it from `src/pluggy/manager.py`?

PR #224, opened 2019-07-21
A few minor fixes

bluetech:
Please see the commit messages.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -777,7 +777,9 @@ using the :py:meth:`pluggy._HookCaller.call_historic()` method:
 
 
     # call with history; no results returned
-    pm.hook.myhook.call_historic(config=config, args=sys.argv, result_callback=callback)
+    pm.hook.myhook.call_historic(
+        kwargs={"config": config, "args": sys.argv}, result_callback=callback
+    )
 
     # ... more of our program ...
 
--- a/src/pluggy/hooks.py
+++ b/src/pluggy/hooks.py
@@ -171,9 +171,9 @@ def varnames(func):
     args, defaults = tuple(spec.args), spec.defaults
     if defaults:
         index = -len(defaults)
-        args, defaults = args[:index], tuple(args[index:])
+        args, kwargs = args[:index], tuple(args[index:])
     else:
-        defaults = ()
+        kwargs = ()
 
     # strip any implicit instance arg
     # pypy3 uses "obj" instead of "self" for default dunder methods
@@ -185,10 +185,10 @@ def varnames(func):
             args = args[1:]
 
     try:
-        cache["_varnames"] = args, defaults
+        cache["_varnames"] = args, kwargs
     except TypeError:
         pass
-    return args, defaults
+    return args, kwargs
 
 
 class _HookRelay(object):
--- a/src/pluggy/hooks.py
+++ b/src/pluggy/hooks.py
@@ -161,7 +161,7 @@ def varnames(func):
         try:
             func = getattr(func, "__call__", func)
         except Exception:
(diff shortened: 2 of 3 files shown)


testing/test_multicall.py
```diff
@@ -1,22 +1,13 @@
...
 from pluggy.hooks import HookImpl
-from pluggy.callers import _multicall, _legacymulticall, _LegacyMultiCall
+from pluggy.callers import _multicall, _legacymulticall
 
 
 hookspec = HookspecMarker("example")
 hookimpl = HookimplMarker("example")
 
 
-def test_uses_copy_of_methods():
```
goodboy: @bluetech you're saying this test was broken for you?

I don't recall seeing that issue but we are planning on removing all of this as per #59 anyway.

docs/index.rst
```diff
@@ -777,7 +777,9 @@ using the :py:meth:`pluggy._HookCaller.call_historic()` method:
 
 
     # call with history; no results returned
-    pm.hook.myhook.call_historic(config=config, args=sys.argv, result_callback=callback)
+    pm.hook.myhook.call_historic(
```
goodboy: Nice catch!

src/pluggy/manager.py
```diff
@@ -135,6 +135,7 @@ def parse_hookimpl_opts(self, plugin, name):
             res = getattr(method, self.project_name + "_impl", None)
         except Exception:
             res = {}
+            normalize_hookimpl_opts(res)
```
goodboy: Hmm maybe we could do the same single conditional block like at line `111` at the end of this method?

src/pluggy/hooks.py
```diff
@@ -171,9 +171,9 @@ def varnames(func):
     args, defaults = tuple(spec.args), spec.defaults
     if defaults:
         index = -len(defaults)
-        args, defaults = args[:index], tuple(args[index:])
+        args, kwargs = args[:index], tuple(args[index:])
```
goodboy: Agreed that this name is better.

testing/test_multicall.py
bluetech: The test has a typo `return` instead of `assert`. If the typo is fixed, it fails. Actually the methods are not copied. Since I think the test is ill conceived and tests deprecated functionality, I opted to just remove it.

goodboy: Sounds good.

src/pluggy/manager.py
bluetech: I can't figure out what "same single conditional block like at line 111" refers to, can you explain more?

goodboy: Ahh sorry my local version of the code is out of date. So basically a block [like this](https://github.com/pytest-dev/pluggy/blob/master/src/pluggy/manager.py#L115) at the end of this method which does the normalization if `res != None`. This might also mean we can remove line `116`? A brief look shows this is the only place that method is used internally so I think we might be ok with that change.

@RonnyPfannschmidt @nicoddemus do we use that function in `pytest` for some reason?

goodboy: [Doesn't look like it](https://github.com/pytest-dev/pytest/search?q=normalize_hookimpl_opts&unscoped_q=normalize_hookimpl_opts). @bluetech I think you should be fine to factor that call into `parse_hookimpl_opts()` and remove it from `116`.

bluetech: There is a test `test_parse_hookimpl_override` which overrides `parse_hookimpl_opts`, returns `{}` and expects it to work. I did not consider this option so the premise of my commit was wrong. I removed it.

PR #227, opened 2019-07-26
Avoid cyclic dependency by merging callers.py into hooks.py

bluetech:
The two modules have a type-level cyclic dependency:
hooks.py uses `_multicall` and `_legacymulticall` from callers.py.
callers.py uses `HookImpl` from hooks.py.

This suggests that they're better off combined.

The `_Result` functionality from callers.py is independent and general, so
move it to its own module _result.py.

Backward compatibility: I did not find any external project which
imports directly from `pluggy.callers`. It exposes two potential import
candidates:
- `HookCallError`: exposed from the top-level, so no reason to reach
  inside.
- `_Result`: it's prefixed and hidden, so no guarantees about that.


src/pluggy/hooks.py
```diff
@@ -360,3 +361,142 @@ def __init__(self, namespace, name, opts):
...
+    call_outcome = _Result.from_call(func)
+    try:
+        wrap_controller.send(call_outcome)
+        _raise_wrapfail(wrap_controller, "has second yield")
+    except StopIteration:
+        pass
+    return call_outcome.get_result()
+
+
+class _LegacyMultiCall(object):
```
nicoddemus: This reminds me, we can drop `_LegacyMultiCall` by now (https://github.com/pytest-dev/pluggy/issues/59).

bluetech: There is an open PR for this #147. Would certainly simplify things to drop it, if it's now possible.

goodboy:
@nicoddemus I would like to review.

One thing I definitely want to maintain in the project is clarity of authorship. Moving big chunks of code around like this can sometimes obscure that.

src/pluggy/hooks.py
```diff
@@ -360,3 +361,142 @@ def __init__(self, namespace, name, opts):
         self.opts = opts
         self.argnames = ["__multicall__"] + list(self.argnames)
         self.warn_on_impl = opts.get("warn_on_impl")
+
```
goodboy: Yeah so this is basically reverting work to break this stuff *out of* `hooks.py` originally (see the history). The idea here was that we want a separate package/module space so we can keep track of and add new caller implementations. 

Imho putting this back into hooks is wrong and doesn't work towards that goal and we'll want to revert this again eventually anyway.

src/pluggy/_result.py
```diff
@@ -0,0 +1,64 @@
+import sys
```
goodboy: Yeah I'm not huge on moving big chunks like this since we lose attribution from the original author in git (unless that's been addressed in recent versions).

I am guilty of this when creating this file (which was an attempt to breakout caller implementations into a separate namespace - see comment below). I really would have liked to have kept @hpk42's original attribution here..

src/pluggy/hooks.py
goodboy: Yes that should land before we start moving big chunks of code around. Let's simplify first then worry about a reorg.

bluetech:
Thanks for the review @goodboy.

> Yes that should land before we start moving big chunks of code around. Let's simplify first then worry about a reorg.

OK, so this is blocked on #147. I'll close this now and maybe revisit when that happens.

PR #229, opened 2019-08-20
tox.ini: make "tox -e py37" etc work

blueyed:


tox.ini
```diff
@@ -9,11 +9,11 @@ commands=
 setenv=
   _PYTEST_SETUP_SKIP_PLUGGY_DEP=1
   coverage: _PLUGGY_TOX_CMD=coverage run -m pytest
+  pytestmaster: _PYTEST_DEP=git+https://github.com/pytest-dev/pytest.git@master
+  pytestfeatures: _PYTEST_DEP=git+https://github.com/pytest-dev/pytest.git@features
 deps=
-  pytestrelease: pytest
```
nicoddemus: Should we remove `-pytestrelease` from `tox.envlist` as well?

blueyed: Thought to keep it for B/C.

blueyed: Removed it.

PR #230, opened 2019-08-20
test_pluginmanager: use pluggy.manager.metadata

blueyed:
This has the version switch in a central place only once.

Ref: https://github.com/pytest-dev/pluggy/pull/223/files#r315505111

### DIFF
--- a/src/pluggy/manager.py
+++ b/src/pluggy/manager.py
@@ -5,9 +5,9 @@ from .hooks import HookImpl, _HookRelay, _HookCaller, normalize_hookimpl_opts
 import warnings
 
 if sys.version_info >= (3, 8):
-    from importlib import metadata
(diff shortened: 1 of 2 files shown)


bluetech:
`manager.metadata` doesn't read too well to me, it's not immediately clear that it refers to `importlib.metadata`.

It will be clearer if, in manager.py, the common name is changed to `importlib_metadata`, that is

```diff
 if sys.version_info >= (3, 8):
-    from importlib import metadata
+    from importlib import metadata as importlib_metadata
 else:
-     import importlib_metadata as metadata
+     import importlib_metadata
```

bluetech:
(And also, do `from pluggy.manager import importlib_metadata` instead of using it qualified).

blueyed:
Done.

PR #232, opened 2019-08-29
Add the output of the example code

kinow:
Hi,

I am trying `pluggy`, and after reading the `README.md` short example I thought I knew what would be the output, but had to run the code to confirm.

This PR adds Python comments to the code block with the results. But perhaps it could go somewhere else if others have any suggestions?

Thanks for pluggy
Bruno

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,6 +61,14 @@ A definitive example
     print(results)
 
 
+Running this directly gets us::
+
+    $ python docs/examples/toy-example.py
+    inside Plugin_2.myhook()
+    inside Plugin_1.myhook()
+    [-1, 3]
+
+
 .. badges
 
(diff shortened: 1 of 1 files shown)


kinow:
Ah, just realized the readthedocs page has the results. Feel free to close if you prefer to have just the code in the `README.md` :+1:

goodboy:
@kinow I'm cool with this - helps with groking the example at a glance.

I wonder can we present it in the readme in a little more of a "shell output"-y way like with [literal blocks](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks)?

kinow:
@goodboy good idea. Copied the text from the readthedocs site and updated the commit :+1:

PR #234, opened 2019-09-10
Preparing release 0.13.0

asottile:


asottile:
primarily to avoid https://gitlab.com/python-devs/importlib_metadata/issues/87

asottile:
due to circular dependencies, the 3.8-dev build for this PR is failing due to `tox -> importlib-metadata`, I believe that's ok to ignore

hroncok:
Unless https://github.com/tox-dev/tox/pull/1368 is released, it will continue to fail.

asottile:
tox 3.14 includes that change

Issue #186, opened 2019-01-05
Docs: some name links and refs are borked


tgoodlet:
Was just re-reading the latest docs and noticed some broken names, namely:
- [add_hookspec()](https://github.com/pytest-dev/pluggy/blame/master/docs/index.rst#L561) (missing a trailing `s`)
- the [call_historic()](https://github.com/pytest-dev/pluggy/blame/master/docs/index.rst#L495) method ref broken due to module path
- a bunch of [private type references](https://github.com/pytest-dev/pluggy/blame/master/docs/index.rst#L362) that aren't exposed (poke @RonnyPfannschmidt since I know you don't want a lot of these *public* per say) / linked properly even though we need to document their usage.

It would probably be good if a third party who didn't write the docs went over them and found more issues.

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

goodboy:
Lol, glad it finally got nailed out 👍🏼

PR #236, opened 2019-09-15
References overhaul

hoefling:
Closing #186.

Running `linkcheck` now yields only one broken link to [tgoodlet](https://github.com/tgoodlet) user account, rest are fixed. Also replaced hardcoded links with `intersphinx` refs where possible. Extended apidoc to a minimal amount of private classes/methods, enough just to fix the existing refs.

The second commit adds some crossrefs between the docstrings and some minor formatting improvements. This one is more a suggestion and I wouldn't mind throwing it away - check whether it really adds a significant amount of readability.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -315,7 +315,7 @@ initial release
 
 .. contributors
 .. _@hpk42: https://github.com/hpk42
-.. _@tgoodlet: https://github.com/tgoodlet
+.. _@tgoodlet: https://github.com/goodboy
 .. _@MichalTHEDUDE: https://github.com/MichalTHEDUDE
 .. _@vodik: https://github.com/vodik
 .. _@RonnyPfannschmidt: https://github.com/RonnyPfannschmidt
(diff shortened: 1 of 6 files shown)


goodboy:
>Running linkcheck now yields only one broken link to tgoodlet user account,

@hoefling that's my old account. I guess just link to my new one?

Other then that this looks mostly good to me though I'll do another run through after others have.

hoefling:
@goodboy Oh, didn't know that - easy fix, then.

goodboy:
@RonnyPfannschmidt @nicoddemus this get's a stamp from me. If you two don't have the time to look through this I think it's ready to land.

hoefling:
@goodboy thanks, glad it was helpful. Sphinx offers more fancy stuff that could be used - for example, I noticed definitions throughout the docs that could be collected under a `.. glossary::` directive and referenced via `:term:`, creating nice hyperlinks to the def paragraph (e.g. for `host`, `plugin`, `hookspec`, `hookimpl` etc). See [stdlib glossary](https://docs.python.org/3/glossary.html) for an example. Or referencing Python keywords via `:keyword:` role, e.g. `yield`, `None`, `True` etc. May be an overkill, though...

PR #248, opened 2020-01-12
ci: add codecov config

blueyed:


blueyed:
Does not show up though..  needs more investigation.

nicoddemus:
It was not enabled for `pluggy`:

image

Enabled, new builds should post coverage results in the PR now.

PR #264, opened 2020-06-04
Allow running `tox -e linting` with any python3

bluetech:
I'm not an expert, but I can't run it otherwise if I only have Python 3.8 installed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,13 +4,11 @@ repos:
     hooks:
     -   id: black
         args: [--safe, --quiet]
-        language_version: python3.6
 -   repo: https://github.com/asottile/blacken-docs
     rev: v0.5.0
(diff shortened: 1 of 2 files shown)


bluetech:
Hmm so I guess it was done because appveyor has an old python3 and black requires 3.6.

I think I'll just work on switching to Github Actions, so I can use same setup as pytest, which works well I think

goodboy:
@bluetech so do we still need this if we're moving to GHA?

bluetech:
> @bluetech so do we still need this if we're moving to GHA?

Yeah, actually I wanted to move to GHA because this PR doesn't work with appveyor. So it actually depends on the appveyor removal (#273).

I usually run `tox -e linting` before submitting PRs, but without this PR it doesn't work unless you have python3.6 installed, which I don't...

PR #266, opened 2020-06-04
Add Github Actions CI for Windows jobs

bluetech:
This can replace Appveyor. I was prompted to this by https://github.com/pytest-dev/pluggy/pull/264#issuecomment-638974445 but also to match what pytest is using.

The idea is to replace Travis as well but that requires some more work.

Github Actions doesn't support Python 3.4, so it's dropped compared to Appveyor, however it is still covered by Travis.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,141 @@
+name: main
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+
+  pull_request:
+    branches:
+      - master
+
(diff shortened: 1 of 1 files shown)


bluetech:
You can see a successful run here: https://github.com/bluetech/pluggy/actions/runs/125108367

bluetech:
I don't have write access to this repository, so if anyone seeing this can merge that would be great :)

Afterwards, I'll send a PR to remove the Appveyor build.

PR #268, opened 2020-06-04
Move `_Result` and friends to new module

goodboy:
Originally proposed (and accepted by moi) in #227.

Relates to:
- fits with discussion in #244 the solution in #257 and further ideas and discussion in #260
- allows for solving a import cycle in #104


goodboy:
Huh, note entirely sure what's up with codecov drop but..

PR #271, opened 2020-06-05
Remove deprecated implprefix support

### LINKED ISSUE #145: Remove implprefix arg to PluginManager
As per discussion in #144 we plan to outright remove the `implprefix` kwarg to `PluginManager` after an appropriate PR has been made to `pytest` implementing the prefix based hook detection/registration.

Further discussion should be made here regarding a possible [strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern) for hook detection outside of the regular explicit marking mechanism.

An initial thought is to instead offer prefix based registration through a `prefix` kwarg to `PluginManager.register()` if this is a feature we'd like to keep.

bluetech:
This has been deprecated for a while, let's remove it in pluggy 1.0.

Refs #116. Fixes #145.

pytest no longer uses it.

tox doesn't use it.

devpi [mentions it](https://github.com/devpi/devpi/search?q=implprefix&unscoped_q=implprefix) with comment "support old plugins, but emit deprecation warnings", does so in a future-compatible manner however (doesn't pass the `implprefix` argument to `PluginManager`, which this PR removes, but sets `_implprefix` attribute instead, which this PR now ignores). devpi uses `pluggy>=0.6.0,<1.0` so it's fine anyway.

### DIFF
--- a/changelog/116.removal.rst
+++ b/changelog/116.removal.rst
@@ -1,0 +1,3 @@
+Remove deprecated ``implprefix`` support.
+Decorate hook implementations using an instance of HookimplMarker instead.
+The deprecation was announced in release ``0.7.0``.
--- a/src/pluggy/manager.py
+++ b/src/pluggy/manager.py
@@ -64,23 +64,13 @@ class PluginManager(object):
     which will subsequently send debug information to the trace helper.
     """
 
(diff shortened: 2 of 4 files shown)


bluetech:
Added "Fixes #145" to the PR description, missed it when I submitted the PR.

PR #274, opened 2020-06-06
Remove Python 2 support

### LINKED ISSUE #272: Drop support for Python 2
Since Python 2 is now officially unsupported upstream, pluggy 1.0 is a good point to drop support for it.

Projects which want to keep support for Python 2 should keep using pluggy 0.6.x series. Pluggy 1.0 will use `python_requires = >=3.5` such that `pip install pluggy` on Python 2 projects will keep getting the 0.6.x series.

The benefits of doing this are:
- Ability to drop Python 2 compatibility code
- Ability to use Python 3-only features, like type annotation syntax, `gen.throw()`, `return` from generator, `yield from` syntax, type annotation syntax and others.
- Less CI/maintenane stuff.

In terms of big projects:

- pytest>=5 is Python 3 only. Old 4.6.x branch supports Python2, uses `pluggy>=0.12,<1.0`, so it's good.

- tox still supports Python 2, though seems to have a 4.0.0 version in development which will drop support for it (I think?). It is requiring pluggy using `pluggy>=0.12.0` , if this is accepted I'll them a PR which changes this to `pluggy>=0.12.0,<1.0.0`.

- devpi seems to have two components which use `pluggy`. The `server` component dropped Python 2. The `client` component supports Python 2 but uses `pluggy>=0.6.0,<1.0` so should be fine.

bluetech:
For pluggy 1.0.

Fixes #272.


bluetech:
There was a `if hasattr(inspect, "signature"):` check that I had missed before, removed now.

goodboy:
@bluetech also needs rebase. Sorry I must have merged in the non-ideal order.

bluetech:
> also needs rebase.

Rebased.

>  Sorry I must have merged in the non-ideal order.

No problem, everything conflicts with everything and I intentionally submit based on current master. Just merge in whatever order and I'll rebase as necessary :)

goodboy:
Gah. I did it again.

bluetech:
Rebased. I also noticed I accidentally added a changelog file in a previous PR to the root instead of changelog/, so just added another commit to fix that here (doesn't seem worth another PR).

setup.py
```diff
@@ -12,10 +12,7 @@
     "Topic :: Utilities",
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
```
hugovk, line 14: Consider making "3 :: Only" explicit, like in pytest?

https://github.com/pytest-dev/pytest/blob/c83801eed0dcfcb14f619dcb5892c9bfca78c573/setup.cfg#L19

```suggestion
    "Programming Language :: Python :: Implementation :: PyPy",
    "Programming Language :: Python :: 3 :: Only",
```

bluetech: Thanks, applied.

PR #279, opened 2020-06-26
Small cleanups

bluetech:
These are some trivial cleanups. See the commit messages for some explanations.

PR #280, opened 2020-06-26
Optimize hook calling a bit

bluetech:
~~*Note: includes PR #279, please ignore the duplicate commits (will rebase once that one is settled).*~~

This PR start to optimize the hook calling path, mostly for the benefit of pytest.

For this pytest file which we use as a useful benchmark of pytest overhead,

```py
import pytest
@pytest.mark.parametrize("x", range(5000))
def test_foo(x): pass
```

Before: `10752102 function calls (10196645 primitive calls) in 9.270 seconds`
After: `10351436 function calls (9896147 primitive calls) in 8.918 seconds`

The main change stems from looking at the stack trace of a hook call. Before this PR it was this (pytest often has several of these nested):

```
File "/pytest/src/_pytest/python.py", line 1561, in runtest
  self.ihook.pytest_pyfunc_call(pyfuncitem=self)
File "/pytest/.tox/venv/lib/python3.8/site-packages/pluggy/hooks.py", line 286, in __call__
  return self._hookexec(self, self.get_hookimpls(), kwargs)
File "/pytest/.tox/venv/lib/python3.8/site-packages/pluggy/manager.py", line 93, in _hookexec
  return self._inner_hookexec(hook, methods, kwargs)
File "/pytest/.tox/venv/lib/python3.8/site-packages/pluggy/manager.py", line 84, in <lambda>
  self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
File "/pytest/.tox/venv/lib/python3.8/site-packages/pluggy/callers.py", line 208, in _multicall
  return outcome.get_result()
File "/pytest/.tox/venv/lib/python3.8/site-packages/pluggy/callers.py", line 80, in get_result
  raise ex[1].with_traceback(ex[2])
File "/pytest/.tox/venv/lib/python3.8/site-packages/pluggy/callers.py", line 187, in _multicall
  res = hook_impl.function(*args)
File "/pytest/src/_pytest/python.py", line 177, in pytest_pyfunc_call
  result = testfunction(**testargs)
```

This PR removes the `<lambda>` frame, and follow up PRs will (try) to remove the `_hookexec` frame and the duplicate `_multicall` frame (which is cosmetic).

### DIFF
--- a/src/pluggy/hooks.py
+++ b/src/pluggy/hooks.py
@@ -248,7 +248,7 @@ class _HookCaller:
         assert not self.is_historic()
 
         if self.spec and self.spec.argnames:
-            notincall = set(self.spec.argnames) - set(kwargs.keys())
+            notincall = set(self.spec.argnames) - kwargs.keys()
             if notincall:
                 warnings.warn(
                     "Argument(s) {} which are declared in the hookspec "
--- a/src/pluggy/hooks.py
+++ b/src/pluggy/hooks.py
@@ -247,14 +247,18 @@ class _HookCaller:
             raise TypeError("hook calling supports only keyword arguments")
         assert not self.is_historic()
 
-        if self.spec and self.spec.argnames:
-            notincall = set(self.spec.argnames) - kwargs.keys()
-            if notincall:
-                warnings.warn(
-                    "Argument(s) {} which are declared in the hookspec "
-                    "can not be found in this hook call".format(tuple(notincall)),
-                    stacklevel=2,
-                )
+        # This is written to avoid expensive operations when not needed.
+        if self.spec:
+            for argname in self.spec.argnames:
+                if argname not in kwargs:
+                    notincall = tuple(set(self.spec.argnames) - kwargs.keys())
+                    warnings.warn(
+                        "Argument(s) {} which are declared in the hookspec "
+                        "can not be found in this hook call".format(notincall),
+                        stacklevel=2,
+                    )
+                    break
+
         return self._hookexec(self, self.get_hookimpls(), kwargs)
 
     def call_historic(self, result_callback=None, kwargs=None):
--- a/src/pluggy/hooks.py
+++ b/src/pluggy/hooks.py
@@ -259,7 +259,11 @@ class _HookCaller:
                     )
                     break
 
-        return self._hookexec(self, self.get_hookimpls(), kwargs)
+            firstresult = self.spec.opts.get("firstresult")
+        else:
+            firstresult = False
+
+        return self._hookexec(self.name, self.get_hookimpls(), kwargs, firstresult)
 
     def call_historic(self, result_callback=None, kwargs=None):
         """Call the hook with given ``kwargs`` for all registered plugins and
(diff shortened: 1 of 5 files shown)


bluetech:
I forgot to mention, it is intended to be reviewed commit-by-commit.

goodboy:
@bluetech oh nice!

I'll try to look at this over the weekend :+1:

bluetech:
Rebased, doesn't depend on other PRs now.

I also remoeved one of the micro-optimization commits, on second thought it's probably not worth it and is distracting.

src/pluggy/hooks.py
```diff
@@ -247,15 +247,23 @@ def __call__(self, *args, **kwargs):
...
-            if notincall:
-                warnings.warn(
-                    "Argument(s) {} which are declared in the hookspec "
-                    "can not be found in this hook call".format(tuple(notincall)),
-                    stacklevel=2,
-                )
-        return self._hookexec(self, self.get_hookimpls(), kwargs)
+        # This is written to avoid expensive operations when not needed.
+        if self.spec:
+            for argname in self.spec.argnames:
```
goodboy, line 252: Is looping through the args everytime hoping for none missing from the call faster then just always checking what's missing using the `set` difference?

Seems like in average case this will be slower (assuming calls are written correctly most of the time)?
I'm not actually sure I can think of a case where this is faster?
I don't think a `for` loop will ever be faster then a `set` difference but I could be wrong?

src/pluggy/hooks.py
```diff
@@ -247,15 +247,23 @@ def __call__(self, *args, **kwargs):
...
+                if argname not in kwargs:
+                    notincall = tuple(set(self.spec.argnames) - kwargs.keys())
+                    warnings.warn(
+                        "Argument(s) {} which are declared in the hookspec "
+                        "can not be found in this hook call".format(notincall),
+                        stacklevel=2,
+                    )
+                    break
+
+            firstresult = self.spec.opts.get("firstresult")
```
goodboy, line 262: This is exactly what I had in mind :)

src/pluggy/hooks.py
bluetech: Here is an unscientific benchmark (Python 3.8, Archlinux), also added a variant which uses `issubset`. Only checks the happy path.

```py
def old(argnames, kwargs):
    if argnames:
        notincall = set(argnames) - kwargs.keys()
        if notincall:
            pass

def old_subset(argnames, kwargs):
    if argnames:
        if not set(argnames).issubset(kwargs.keys()):
            pass

def new(argnames, kwargs):
    for argname in argnames:
        if argname not in kwargs:
            break

import timeit
kwargs = {'a': 0, 'b': 1, 'c': 2, 'd': 3, 'e': 4}
argnames = list(kwargs)
print("old:       ", timeit.timeit("old(argnames, kwargs)", "from __main__ import old, argnames, kwargs"))
print("old_subset:", timeit.timeit("old_subset(argnames, kwargs)", "from __main__ import old_subset, argnames, kwargs"))
print("new:       ", timeit.timeit("new(argnames, kwargs)", "from __main__ import new, argnames, kwargs"))
```

Output:

```
old:        0.7920419139554724
old_subset: 0.6716860989108682
new:        0.2587350399699062
```

goodboy: > (Python 3.8, Archlinux)

You're of my kind :smile_cat: 

>new:        0.2587350399699062

So slick; I guess `for` the win :surfing_man:

PR #287, opened 2021-03-27
fix: dictionary changed size during iteration iterating on importlib_metadata.distributions()

sbrugman:
Simple fix for this error: https://github.com/python/importlib_metadata/issues/293
(similar to https://github.com/pytest-dev/py/issues/218)


bluetech (approved):
Thanks! LGTM.

CI failures are unrelated.

PR #290, opened 2021-04-02
ci: move remaining travis to github actions, remove travis

bluetech:
Travis is shutting down or something similar in a few weeks. Currently the following jobs are run on travis:

- Testing for Linux
- Coverage reporting
- Deploy
- IRC notifications

Other than the IRC, which I don't think github actions supports, and don't know if still desirable, this moves everything to github actions and deletes the travis file.


bluetech:
So the testing and coverage seem to work.

I didn't test the deploy part, but I think it should work once the pypi token is added as a secret to the repo. One of the [PyPI maintainers](https://pypi.org/project/pluggy/) will need to [issue one](https://pypi.org/help/#apitoken) and [add it](https://github.com/pytest-dev/pluggy/settings/secrets/actions) as a repository secret with name `pypi_token`.

goodboy:
@bluetech looks like I have access.

Will come back with it.

goodboy:
Done. Not sure if there's a way we can test without an actual deploy.

Man, we really got fancy with the CD stuff eh 🏄🏼

goodboy (approved):

RonnyPfannschmidt (approved):

bluetech:
Thanks for the reviews, and thanks @goodboy for adding the token, looks good. Yes, I think we'll only find out if it works when we try to deploy :)

PR #314, opened 2021-08-09
Prefix all internal modules with _ to clarify they're private

### LINKED ISSUE #313: Prefix all internal modules with _ to clarify they're private
Some are already prefixed, but some aren't.

This clarifies that pluggy's only public API is what's exported from the top-level `pluggy` module.

This will allow internal code to be nicer, not having to defensively prefix internal types and functions with _.

I will send a PR doing this for the upcoming 1.0 release.

bluetech:
Some were already prefixed, but some weren't.

This clarifies that pluggy's only public API is what's exported from the
top-level `pluggy` module.

This will allow internal code to be nicer, not having to defensively
prefix internal types and functions with _.

Fixes #313.


nicoddemus (approved):
Awesome! 🚀

RonnyPfannschmidt (approved):

PR #318, opened 2021-08-14
Use refs instead of direct urls to link to pytest docs, fix internal crossrefs

### LINKED ISSUE #317: use intersphinx for references to pytest docs

hoefling:
Closes #317.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -349,7 +349,7 @@ will be executed *first* or *last* respectively in the hook call loop:
     # load a plugin defined on a class
     pm.register(SomeOtherPlugin())
 
-For another example see the `hook function ordering`_ section of the
+For another example see the :ref:`pytest:plugin-hookorder` section of the
 ``pytest`` docs.
 
 .. note::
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):
lovely, before i even got to check how we do it you resolved it, thanks!

hoefling:
@RonnyPfannschmidt gladly - we have internal tools for housekeeping that do that semi-automatically. In a nutshell, you generate a human-readable inventory by running `python -m sphinx.ext.intersphinx https://docs.pytest.org/en/latest/objects.inv > pytest.intersphinx` and look for the path that matches the URL. E.g. for `https://docs.pytest.org/en/latest/_modules/_pytest/hookspec.html#pytest_cmdline_main`, the correspondent line is
```
	_pytest.hookspec.pytest_cmdline_main     reference/reference.html#pytest.hookspec.pytest_cmdline_main
```
(the URLs probably have changed on `pytest` since the last update of `pluggy` docs). The line is found under the section `py:function`, so a replacement is ``` :func:`_pytest.hookspec.pytest_cmdline_main` ```. Where the `func` comes from - see [Cross-referencing Python objects](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects) for roles listing in Python domain. Etc.

nicoddemus (approved):
Thanks @hoefling!

I have built the docs and checked them locally, works perfectly.

(I've enabled PR builds in readthedocs, so docs will be built for each PR from now on).

PR #344, opened 2022-01-11
Add py.typed marker file

layday:
To signal to type checkers that pluggy is typed now.


nicoddemus (approved):

bluetech:
I intentionally didn't publish the types because I think they're not ready yet. There at least needs to be an audit of the API similar to https://github.com/pytest-dev/pytest/issues/7469 before then.

PR #352, opened 2022-02-21
Remove manual kwargs-only check in call

### LINKED ISSUE #350: Remove manual kwargs-only check?
A little trivial issue.

The hookcaller `__call__` (e.g. `pm.hook.myhook(...)`) requires keyword-only arguments. This is currently checked manually:

https://github.com/pytest-dev/pluggy/blob/172ca9f59e8f1029ec734647b38ce4e2c4d7dba2/src/pluggy/_hooks.py#L384-L386

and provides a nice exception message (though it could be nicer...).

An alternative is to make the function itself kwargs-only:

```py
    def __call__(self, **kwargs: object) -> Any:
```

then Python itself checks it. This provides a worse exception message: `_HookCaller.__call__() takes 1 positional argument but 2 were given` (it's still possible to find the line which has the problem from the stacktrace).

On the other hand, the Python kwargs-only approach has the better typing (i.e. mypy will be able to catch bad usages), and removes the slight overhead of the check (likely not significant though).

I'm inclined to make the change but wonder if others agree.

bluetech:
Fix #350.

### DIFF
--- a/src/pluggy/_hooks.py
+++ b/src/pluggy/_hooks.py
@@ -381,9 +381,7 @@ class _HookCaller:
                     )
                     break
 
-    def __call__(self, *args: object, **kwargs: object) -> Any:
-        if args:
-            raise TypeError("hook calling supports only keyword arguments")
+    def __call__(self, **kwargs: object) -> Any:
         assert (
             not self.is_historic()
         ), "Cannot directly call a historic hook - use call_historic instead."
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):
as type-checking is beginning to be the standard, the type-checker can probvide the details

i wonder if it may be sensible to have a type checking plug-in for mypy to ease some details
(and how hard it would be)

bluetech:
> i wonder if it may be sensible to have a type checking plug-in for mypy to ease some details (and how hard it would be)

It would be very helpful, though I've never written one so not sure how feasible it is.

PR #388, opened 2023-06-11
result: use exception instead excinfo triplet

bluetech:
The excinfo triplet is a legacy concept, these days the exception instance itself carries the necessary info.

Switch the internal implementation to store just the exception instead of the triplet, and add an `exception` property to access it without the excinfo annoyance.


RonnyPfannschmidt (approved):
Better to Land this first than to wait for my async change

PR #411, opened 2023-06-20
Make new-style wrappers use explicit `wrapper=True`

### LINKED ISSUE #405: Require new-style hooks to also be marked as hook-wrappers
As discussed in #403, implicitly considering hooks implemented as generators to be hook-wrappers broke some downstream projects, notably `conda`.

@bluetech:

> If the yanking ends up being good enough, I'll work on a pluggy 1.2.0 which reworks the feature (probably using an explicit `wrapper=True`), otherwise I can do a quick 1.1.1 release which just reverts the feature.

> For now I'll assume the yanks are good, but let me know if not.

_Originally posted by @bluetech in https://github.com/pytest-dev/pluggy/issues/403#issuecomment-1597449877_

bluetech:
Previously new-style wrappers were implied by the function being a generator, however that broke a use case of a non-wrapper impl being a generator as a way to return an Iterator result. This is legitimate, and used at least by conda (see #403).

Therefore, change to require an explicit `wrapper=True`. Also adds a test for the iterator-returning case.

Fix #405.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -366,19 +366,25 @@ Wrappers
 .. note::
     This section describes "new-style hook wrappers", which were added in Pluggy
     1.1. For earlier versions, see the "old-style hook wrappers" section below.
-    The two styles are fully interoperable.
 
-A *hookimpl* can be a generator function, which indicates that the function will
-be called to *wrap* (or surround) all other normal *hookimpl* calls. A *hook
-wrapper* can thus execute some code ahead and after the execution of all
-corresponding non-wrappper *hookimpls*.
+    New-style hooks wrappers are declared with ``wrapper=True``, while
+    old-style hook wrappers are declared with ``hookwrapper=True``.
+
+    The two styles are fully interoperable between plugins using different
+    styles. However within the same plugin we recommend using only one style,
+    for consistency.
+
+A *hookimpl* can be marked with the ``"wrapper"`` option, which indicates
+that the function will be called to *wrap* (or surround) all other normal
+*hookimpl* calls. A *hook wrapper* can thus execute some code ahead and after the
+execution of all corresponding non-wrappper *hookimpls*.
 
 Much in the same way as a :py:func:`@contextlib.contextmanager <python:contextlib.contextmanager>`,
 *hook wrappers* must be implemented as generator function with a single ``yield`` in its body:
 
 .. code-block:: python
 
-    @hookimpl
+    @hookimpl(wrapper=True)
     def setup_project(config, args):
         """Wrap calls to ``setup_project()`` implementations which
         should return json encoded config options.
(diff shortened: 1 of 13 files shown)


docs/index.rst
```diff
@@ -368,17 +368,17 @@ Wrappers
     1.1. For earlier versions, see the "old-style hook wrappers" section below.
     The two styles are fully interoperable.
 
-A *hookimpl* can be a generator function, which indicates that the function will
-be called to *wrap* (or surround) all other normal *hookimpl* calls. A *hook
-wrapper* can thus execute some code ahead and after the execution of all
-corresponding non-wrappper *hookimpls*.
+A *hookimpl* can be marked with the ``"hookwrapper"`` option, which indicates
```
nicoddemus, line 371:
```suggestion
A *hookimpl* can be marked with the ``"wrapper"`` option, which indicates
```

docs/index.rst
```diff
@@ -368,17 +368,17 @@ Wrappers
     1.1. For earlier versions, see the "old-style hook wrappers" section below.
     The two styles are fully interoperable.
```
nicoddemus, line 369: We should mention the distinct option names next to each other, as this will help users search for them and be told of the difference explicitly.

```suggestion
    New-style hooks wrappers are declared with ``wrapper=True``, while old-style hook wrappers are declared with ``hookwrapper=True`` (for backward compatibility).

    The two styles are fully interoperable between plugins using different styles. However within the same plugin we recommend using only one style, for consistency.
```

testing/test_invocations.py
```diff
@@ -287,3 +290,39 @@ def hello(self, arg):
...
+
+
+def test_non_wrapper_generator(pm: PluginManager) -> None:
+    """A hookimpl can be a generator without being a wrapper,
+    meaning it returns an iterator result."""
+
+    class Api:
+        @hookspec
+        def hello(self) -> Iterator[int]:
+            raise NotImplementedError()
```
nicoddemus, line 302: We could add `raise NotImplementedError` to `.coveragerc` as a line to be excluded from coverage.

testing/test_pluginmanager.py
```diff
@@ -156,6 +156,21 @@ def he_method1(self):
     assert excinfo.value.plugin is plugin
 
 
+def test_register_both_wrapper_and_hookwrapper(he_pm: PluginManager) -> None:
+    class hello:
+        @hookimpl(wrapper=True, hookwrapper=True)
+        def he_method1(self):
+            yield
```
nicoddemus, line 163:
```suggestion
            yield  # pragma: no cover
```

nicoddemus (approved):
Great work! Left a few comments.

docs/index.rst
bluetech: Thanks! I'll apply this, except for the "(for backward compatibility)" part, I think it is potentially confusing.

nicoddemus: Ahh right, might give the impression we are deprecating old-style hook wrappers, which is not the case.

PR #412, opened 2023-06-21
Preparing release 1.2.0

bluetech:
I've ran the downstream tests and they seem good AFAICT. So I think now it will go smoothly :)


RonnyPfannschmidt (approved):
Lovely, great work, thanks!

PR #423, opened 2023-07-20
ensure the type hints support the writer unsetting in tagtracer

hoefling:
Reason: `pytest` uses `TagTracer.setwriter(None)` when unsetting the tracer in https://github.com/pytest-dev/pytest/blob/1de0923e83e3258709559d86d142adaab65a42d8/src/_pytest/helpconfig.py#L126. This PR adjusts the typing of `TagTracer.setwriter` to support this case.


bluetech (approved):
Thanks

PR #430, opened 2023-08-13
More steps towards exporting our typings

bluetech:
See #428. I believe this handles all of the remaining items.

After this PR, I'll send a final PR which exports the typings and adds a changelog with all of the changes.


bluetech:
The "final" API Reference: https://pluggy--430.org.readthedocs.build/en/430/api_reference.html

nicoddemus (approved):
LGTM

PR #435, opened 2023-08-26
Preparing release 1.3.0

bluetech:
Notes:

- Downstream testing seems OK
- I also did mypy downstream testing:
    - pytest: https://github.com/pytest-dev/pytest/pull/11353

    - tox: has a few new "useless type ignore" warnings, expected

    - devpi: has a new mypy error due to using `warn_on_impl="a string"` instead of `warn_on_impl=Warning("a warning")`. I'm not really sure if the string one even works, but in any case I think it's better to let devpi change to `Warning` here instead of adding support for str.

    - The rest seem good.


nicoddemus (approved):

bluetech:
Pluggy 1.3.0 is released, thanks for the reviews as always @nicoddemus.

nicoddemus:
Awesome, thanks for taking care of the release! 👍

PR #446, opened 2023-09-22
add test coverage for weird case of plugin with multiple impls of same spec

wahuneke:
These tests are attempting to find @RonnyPfannschmidt 's theoretical issue described in issue #431 . Conversation in 431 is ongoing.  This PR is here for reference.  It is a DRAFT.

**NOTE: these tests currently pass.  They do not detect any new issues.**

### DIFF
--- a/testing/test_tracer.py
+++ b/testing/test_tracer.py
@@ -1,6 +1,13 @@
 import pytest
 
+from pluggy import HookimplMarker
+from pluggy import HookspecMarker
+from pluggy import PluginManager
 from pluggy._tracing import TagTracer
+
+
+hookspec = HookspecMarker("example")
+hookimpl = HookimplMarker("example")
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (approved):
well done, those tests verify that my assessment from just/only looking at the code was wrong and now we have a demonstration of the correctness for those cases

RonnyPfannschmidt:
Rebased onto current main and updated for merge:

- **Dropped the `test_hook_multi_impl` commit** — the behaviour it tested (register/unregister with multi-impl plugins, `get_hookcallers` return count) is now fully covered by `test_unregister_plugin_with_multi_hookimpls` and `test_get_hookcallers_no_duplicates` added in #646, which also fixed the underlying bug (#431) that this PR was originally investigating.

- **Kept the tracer tests** (`test_plugin_tracing`, `test_dbl_plugin_tracing`) — these are new coverage for hook tracing with multi-impl plugins, not covered elsewhere.

- Fixed `List[str]` → `list[str]` and added a missing blank line to pass pre-commit (ruff, flake8, mypy).

PR #447, opened 2023-09-23
add force_not_a_hook() and unit test

wahuneke:
This DRAFT pull request gets the ball rolling on a possible solution for issue #249


RonnyPfannschmidt:
in the issue we discovered that its not necessary anymore as plain pluggy no longer does this

PR #452, opened 2023-10-03
add test cases to improve coverage stat

wahuneke:


wahuneke:
I'm not sure why it is unable to run and update the coverage stats.  Perhaps this is a known issue to someone else here on the project....

In any case, my local runs showed coverage improving.  For the precise improvement amount (coverage before and after) I'd need to see the codecov.io report.  For now, let's just say: this PR fills in some gaps but one or two still remain (e.g. coverage for pypy-only branch)

PR #460, opened 2023-12-13
Issue 424/add saferepr

anusha-shekhar:


testing/test_hookcaller.py
```diff
@@ -448,3 +449,24 @@ def conflict(self) -> None:
...
+) -> None:
+    @addmeth()
+    def he_method1() -> None:
+        pass
+
+    @addmeth()
+    def he_method2() -> None:
+        # Intentional error to make the repr fail
+        raise ValueError("Intentional error in he_method2")
+
```
RonnyPfannschmidt, line 465: This is not an error unable for the particular test

anusha-shekhar: I am not sure I understand the comment here, I am just trying to mock a random error so that we can test the saferepr implementation.

anusha-shekhar:
Hi! It seems that some of the code coverage pipeline is inaccurate. It is counting empty lines as requiring test coverage (please see screenshot provided)

Screenshot 2023-12-13 at 11 36 23 PM

src/pluggy/_hooks.py
```diff
@@ -456,7 +457,7 @@ def _add_hookimpl(self, hookimpl: HookImpl) -> None:
             self._hookimpls.insert(i + 1, hookimpl)
 
     def __repr__(self) -> str:
-        return f"<HookCaller {self.name!r}>"
```
RonnyPfannschmidt, line 459: Why blindly replace all usages?

testing/test_hookcaller.py
```diff
@@ -448,3 +452,64 @@ def conflict(self) -> None:
...
+        "tryfirst": False,
+        "trylast": False,
+        "specname": "",
+    }
+
+    # Initialize HookImpl
+    hook_impl = HookImpl(plugin, plugin_name, example_function, hook_impl_opts)
+
+    # Verify attributes are set correctly
+    assert hook_impl.function == example_function
```
RonnyPfannschmidt, line 478: that type of test adds no value as far as  can tell - lets keep experiments out of the codebase

testing/test_hookcaller.py
```diff
@@ -448,3 +452,64 @@ def conflict(self) -> None:
...
+        "trylast": False,
+        "specname": "",
+    }
+
+    # Initialize HookImpl
+    hook_impl = HookImpl(plugin, plugin_name, example_function, hook_impl_opts)
+
+    # Verify __repr__ method
+    expected_repr = (
+        f"<HookImpl plugin_name={saferepr(plugin_name)}, " f"plugin={saferepr(plugin)}>"
```
RonnyPfannschmidt, line 513: we shouldnt add safe-repr all over the place for everything

RonnyPfannschmidt (changes requested):
as far as i can tell this pull request blindly replaces calls, but does not address the casting issue in the tracing code at all

based on the code as im looking at now, a simplified try/except based "safe str" is needed


also a test using tracing that traces a broken object as input is needed,

PR #464, opened 2023-12-21
Warn when old-style hookwrapper raises during teardown

### LINKED ISSUE #463: Warn when old-style hookwrapper raises during teardown
Old-style hookwrapper raising during teardown causes subsequent teardowns to be deferred until GC (see #244). We've fixed this in pluggy 1.2 using new-style wrappers or `Result.force_exception`.

If we detect the situation, let's warn about it. See for example https://github.com/pytest-dev/pytest/issues/11714 where a warning would have helped.

bluetech:
Fix #463.

### DIFF
--- a/changelog/463.feature.rst
+++ b/changelog/463.feature.rst
@@ -1,0 +1,2 @@
+A warning :class:`~pluggy.PluggyTeardownRaisedWarning` is now issued when an old-style hookwrapper raises an exception during teardown.
+See the warning documentation for more details.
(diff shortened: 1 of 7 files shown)


bluetech:
Before this can be merged, I want to check how this affects some pluggy users. If this issue is widespread we might want to handle it differently.

bluetech:
Did downstream testing, seems fine. Though the problems are more likely to occur in 3rd-party plugins of these downstreams, which I didn't check.

I also verified that the warning is issued on the pytest-mpl test suite.

nicoddemus (approved):
LGTM!

PR #473, opened 2024-01-20
hooks: fix plugins registering other plugins in a hook

### LINKED ISSUE #438: pytest won't run root conftest.py's pytest_configure in special circumstances
Wanted to cross post https://github.com/pytest-dev/pytest/issues/11365 here, in case it might be on the pluggy side:

> Consider the following project: https://github.com/Redoubts/_pytest_bug_report
> 
> There's a root level conftest.py, and 3 plugins (two loaded in the conftest, and a third loaded by plugin 2). All define a `pytest_configure` in some way.
> 
> with `platform darwin -- Python 3.11.4, pytest-7.4.0, pluggy-1.0.0`, we see
> 
> ```
> % pytest test1.py
> mypackage.plugin2 Configured
> mypackage.plugin3 Configured
> conftest Configured
> mypackage.plugin Configured
> ```
> 
> Which is somewhat expected, each `pytest_configure` gets called.
> 
> If you update pluggy to 1.2+, you see
> 
> ```
> % pytest test1.py           
> mypackage.plugin2 Configured
> mypackage.plugin3 Configured
> mypackage.plugin Configured
> mypackage.plugin3 Configured
> ============================================================================================ test session starts =============================================================================================
> platform darwin -- Python 3.11.4, pytest-7.4.0, pluggy-1.2.0
> ...
> ```
> 
> Where the third plugin's `pytest_configure` is called twice, but the root `pytest_configure` is never called.
> 
> I think this is more specifically due to `mypackage/plugin3.py`'s `@pytest.hookimpl` wrapper. Maybe it's ill advised to do that, but I found this bizarre scenario when testing package updates in a much larger codebase.

bluetech:
when the other plugins implement the same hook itself.

Fix #438.
Regressed in 63b7e908b4b22c30d86cd2cff240b3b7aa6da596 - pluggy 1.1.0.

Went with the simple solution described in the issue for now, basically undoing most of the rationale for 63b7e908b4b22c30d86cd2cf (though I think it's still better to have a single `_hookimpls` list).

### DIFF
--- a/src/pluggy/_hooks.py
+++ b/src/pluggy/_hooks.py
@@ -497,7 +497,8 @@ class HookCaller:
         ), "Cannot directly call a historic hook - use call_historic instead."
         self._verify_all_args_are_provided(kwargs)
         firstresult = self.spec.opts.get("firstresult", False) if self.spec else False
-        return self._hookexec(self.name, self._hookimpls, kwargs, firstresult)
+        # Copy because plugins may register other plugins during iteration (#438).
+        return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
 
     def call_historic(
         self,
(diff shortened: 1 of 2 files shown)


nicoddemus (approved):
LGTM!

Should we add a CHANGELOG entry too?

RonnyPfannschmidt (approved):

bluetech:
> Should we add a CHANGELOG entry too?

I will update the changelogs before doing a release. I will do a release in the next few days.

PR #479, opened 2024-02-06
Update index.rst

beedub:
Fixing what I believe are typos in the docs


nicoddemus (approved):
Indeed they are, thanks!

PR #495, opened 2024-04-18
Add support deprecating hook parameters

### LINKED ISSUE #178: Deprecate hook parameters
We already have support for deprecating hooks, where a warning is issued if any plugin implements a deprecated hook.

Recently in pytest 3.8 we introduced a new hook, [pytest_warning_captured(warning_message, when, item)](https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_warning_captured).

Problem is that only later I realized that `item` is incompatible with xdist (https://github.com/pytest-dev/pytest-xdist/issues/341) because we can't transfer `pytest.Item` objects between workers and master. 😓 

I would like then to deprecate the `item` parameter by issuing a warning if a hook implementation uses it.

Given our current deprecation support of hook impls:

```python
@hookspec(historic=True, warn_on_impl=DeprecationWarning("pytest_namespace() is deprecated"))
def pytest_namespace():
    ...
```

I propose:

```python
@hookspec(historic=True, warn_on_param_impl={'item': DeprecationWarning('item deprecated because of reasons'})
def pytest_warning_captured(warning_message, item, when):
    ...
```

What do you guys think?

bluetech:
pytest has some hooks with deprecated parameters [see here](https://docs.pytest.org/en/8.0.x/deprecations.html#py-path-local-arguments-for-hooks-replaced-with-pathlib-path). We were able to apply some hack to raise a deprecation for *callers* passing these parameters, but for deprecating *hookimpls* which requests these parameters, we need support from pluggy.

This adds support for doing this:

```py
@hookspec(
    warn_on_impl_args={
        "lousy_arg": DeprecationWarning(
            "The lousy_arg parameter of refreshed_hook is deprecated and will be removed soon; "
            "use awesome_arg instead"
        ),
    },
)
def refreshed_hook(lousy_arg, awesome_arg):
    pass
```

I have verified that this works for pytest.

Fix #178.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -659,10 +659,33 @@ If a hookspec specifies a ``warn_on_impl``, pluggy will trigger it for any plugi
 .. code-block:: python
 
     @hookspec(
-        warn_on_impl=DeprecationWarning("oldhook is deprecated and will be removed soon")
+        warn_on_impl=DeprecationWarning("old_hook is deprecated and will be removed soon")
     )
-    def oldhook():
+    def old_hook():
         pass
+
+
+If you don't want to deprecate implementing the entire hook, but just specific
+parameters of it, you can specify ``warn_on_impl_args``, a dict mapping
+parameter names to warnings. The warnings will trigger whenever any plugin
+implements the hook requesting one of the specified parameters.
+
+.. code-block:: python
+
+    @hookspec(
+        warn_on_impl_args={
+            "lousy_arg": DeprecationWarning(
+                "The lousy_arg parameter of refreshed_hook is deprecated and will be removed soon; "
+                "use awesome_arg instead"
+            ),
+        },
+    )
+    def refreshed_hook(lousy_arg, awesome_arg):
+        pass
+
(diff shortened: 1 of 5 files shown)


docs/index.rst
```diff
@@ -659,11 +659,31 @@ If a hookspec specifies a ``warn_on_impl``, pluggy will trigger it for any plugi
...
-    def oldhook():
+    def old_hook():
         pass
 
+
+If you don't want to deprecate implementing the entire hook, but just specific
+parameters of it, you can specify ``warn_on_impl_args``, a dict mapping
+parameter names to warnings. The warnings will trigger whenever any plugin
+implements the hook requesting one of the specified parameters.
+
```
nicoddemus, line 672:
```suggestion

.. versionadded:: 1.5

```

nicoddemus (approved):
LGTM, thanks!

RonnyPfannschmidt (approved):
this looks good - im hoping we can expand it to ensure forward/backward compatibility with hook callers at some point

PR #496, opened 2024-04-19
ci: replace upload-coverage script with codecov github action

bluetech:
Same as what pytest does, hopefully will fix coverage upload failures.


RonnyPfannschmidt (approved):
:bar_chart:

PR #556, opened 2024-12-07
Drop Python 3.8 support

nicoddemus:
Also run pre-commit (with pyupgrade) on all files.

### DIFF
--- a/src/pluggy/_hooks.py
+++ b/src/pluggy/_hooks.py
@@ -4,21 +4,19 @@ Internal hook annotation, representation and calling machinery.
 
 from __future__ import annotations
 
+from collections.abc import Generator
+from collections.abc import Mapping
+from collections.abc import Sequence
+from collections.abc import Set
 import inspect
 import sys
 from types import ModuleType
-from typing import AbstractSet
 from typing import Any
 from typing import Callable
 from typing import Final
 from typing import final
-from typing import Generator
-from typing import List
-from typing import Mapping
 from typing import Optional
 from typing import overload
-from typing import Sequence
-from typing import Tuple
 from typing import TYPE_CHECKING
 from typing import TypedDict
 from typing import TypeVar
@@ -34,7 +32,7 @@ _Namespace = Union[ModuleType, type]
 _Plugin = object
 _HookExec = Callable[
     [str, Sequence["HookImpl"], Mapping[str, object], bool],
-    Union[object, List[object]],
+    Union[object, list[object]],
 ]
 _HookImplFunction = Callable[..., Union[_T, Generator[None, Result[_T], None]]]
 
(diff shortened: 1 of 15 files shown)


Pierre-Sassoulas (commented):
I tried to relaunch the coverage, but it seems a genuine drop in coverage. Not sure what changed to have a 5% drop. Maybe code that is only runnable in python 3.8 that we could remove ?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist=docs,py{38,39,310,311,312,313,py3},py{38}-pytest{main}
+envlist=docs,py{39,310,311,312,313,py3},py{39}-pytest{main}
```
RonnyPfannschmidt, line 2: Do we remember why main Is in brackets?

RonnyPfannschmidt (approved):

src/pluggy/_hooks.py
```diff
@@ -608,7 +606,7 @@ class _SubsetHookCaller(HookCaller):
         "_remove_plugins",
     )
 
-    def __init__(self, orig: HookCaller, remove_plugins: AbstractSet[_Plugin]) -> None:
+    def __init__(self, orig: HookCaller, remove_plugins: Set[_Plugin]) -> None:
```
RonnyPfannschmidt, line 609: This one seems incorrect at first glance

tox.ini
nicoddemus: No, I also found that odd. Will remove it.

src/pluggy/_hooks.py
nicoddemus: mypy or ruff (I cannot recally which one) complained and suggested using `Set` from `collections.abc` instead of deprecated `AbstractSet`.

nicoddemus:
> I tried to relaunch the coverage, but it seems a genuine drop in coverage. Not sure what changed to have a 5% drop. Maybe code that is only runnable in python 3.8 that we could remove ?

I can't figure out either... it might be just me, but I can't decipher the codecov report:

image

Blank lines are not covered? That does not make sense to me.

tox.ini
nicoddemus: Done.

src/pluggy/_hooks.py
nicoddemus: Yeah seems correct:

https://docs.python.org/3/library/typing.html#typing.AbstractSet

image

nicoddemus:
Anything else other than the codecov issue folks? Unfortunately I cannot really understand what is missing coverage from that report...

Pierre-Sassoulas (approved):
After examination, the coverage issue seems to be a bug from codecov (only blank lines or comment are not covered). I think we should merge with administrator privileges and bypass that. Everything else LGTM.

PR #627, opened 2025-11-19
Unicode encoding error

Ghanchu:
Fixing an issue with Pluggy where surrogate escape characters are not encoded correctly, causing a crash of Pluggy hooks. From issue [13750](https://github.com/pytest-dev/pytest/issues/13750) in Pytest.

### DIFF
--- a/testing/test_tracer.py
+++ b/testing/test_tracer.py
@@ -75,3 +75,14 @@ def test_setprocessor(rootlogger: TagTracer) -> None:
     log2("seen")
     tags, args = l2[0]
     assert args == ("seen",)
+
+
+def test_unicode_surrogate_handling(rootlogger: TagTracer) -> None:
+    out: list[str] = []
+    rootlogger.setwriter(out.append)
+    log = rootlogger.get("pytest")
+    s = "hello \ud800 world"
+    log(s)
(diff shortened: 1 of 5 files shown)


testing/test_tracer.py
```diff
@@ -75,3 +75,38 @@ def test_setprocessor(rootlogger: TagTracer) -> None:
...
+    bad = b"\xed\xa0\x80".decode("utf-8", "surrogatepass")
+
+    log(bad)
+
+    assert len(out) == 1
+    assert "\ud800" not in out[0]
+    assert "?" in out[0]
+
+
+def test_unicode_surrogate_handling_normal(rootlogger: TagTracer) -> None:
```
Pierre-Sassoulas: This test we can remove, it's implicitely tested everywhere else ;)

Pierre-Sassoulas:
Thank you for the PR @Ghanchu !

RonnyPfannschmidt:
if hte repr combinations are used as such i expect pytest logtraces to use double repr in some parts as well as quoting way too many things

RonnyPfannschmidt:
closing as stale for now - we may reiterate where exactly to put repr in a new issue/pr as the current state accidentially kills the ux

PR #659, opened 2026-03-24
create an optimized shortcut if `__code__` is seen and callable is a routine

Vizonex:
While in the middle of trying to create my own asynchronous callback library I came up with a more optimized approach for extracting function arguments and I wanted to see if this could be a little bit faster here since construction of `inspect.signature` can be a little slow depending on the circumstances.


RonnyPfannschmidt:
This is accidentally a dupe of #632

Vizonex:
@RonnyPfannschmidt Would you like a reviewer for your version of this PR?

PR #666, opened 2026-04-27
Handle broken repr in tracing

### LINKED ISSUE #424: Avoid erroring from tracing
A user reported a weird condition in which pluggy tracing was making their code fail after we enabled tracing because one of the local objects didn't have a working `__repr__` (a scikit-learn object from a very old version). Details: https://github.com/kedro-org/kedro/issues/2630

We gave the user a workaround and also it's reasonable to expect that objects have a working `__repr__`, but it would be nice if there was a generous `try ... except` around the tracing messages so that errors there don't bubble up. Essentially the same that happens with logging:

```
>>> try:
...   logging.warning("Will fail: %d", "a")
...   logging.warning("But the show must go on")
... except:
...   logging.error("Oh no")
... else:
...   logging.warning("Everything is fine")
... 
--- Logging error ---
  File "/opt/homebrew/Cellar/python@3.11/3.11.4/Frameworks/Python.framework/Versions/3.11/lib/python3.11/logging/__init__.py", line 1110, in emit
    msg = self.format(record)
          ^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.4/Frameworks/Python.framework/Versions/3.11/lib/python3.11/logging/__init__.py", line 953, in format
    return fmt.format(record)
           ^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.4/Frameworks/Python.framework/Versions/3.11/lib/python3.11/logging/__init__.py", line 687, in format
    record.message = record.getMessage()
                     ^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.4/Frameworks/Python.framework/Versions/3.11/lib/python3.11/logging/__init__.py", line 377, in getMessage
    msg = msg % self.args
          ~~~~^~~~~~~~~~~
TypeError: %d format: a real number is required, not str
Call stack:
  File "<stdin>", line 2, in <module>
Message: 'Will fail: %d'
Arguments: ('a',)
WARNING:root:But the show must go on
WARNING:root:Everything is fine
```

ShipItAndPray:
## Summary
- guard tracing stringification so hook tracing does not break hook execution when traced objects have a broken `__repr__`/`__str__`
- keep the fallback output informative by using the default object repr and the original exception
- add a regression test covering broken repr values in traced hook kwargs

## Testing
- `uv run --python 3.12 pytest testing/test_pluginmanager.py -k broken_repr -q`
- `uv run --python 3.12 pytest testing/test_pluginmanager.py -q`

Closes #424.

### DIFF
--- a/src/pluggy/_tracing.py
+++ b/src/pluggy/_tracing.py
@@ -13,6 +13,13 @@ _Writer = Callable[[str], object]
 _Processor = Callable[[tuple[str, ...], tuple[Any, ...]], object]
 
 
+def _safe_str(obj: object) -> str:
+    try:
+        return str(obj)
+    except Exception as exc:
+        return f"{object.__repr__(obj)} (str failed: {type(exc).__name__}: {exc})"
+
+
 class TagTracer:
     def __init__(self) -> None:
         self._tags2proc: dict[tuple[str, ...], _Processor] = {}
@@ -29,13 +36,13 @@ class TagTracer:
         else:
             extra = {}
 
-        content = " ".join(map(str, args))
+        content = " ".join(_safe_str(arg) for arg in args)
         indent = "  " * self.indent
 
         lines = [f"{indent}{content} [{':'.join(tags)}]\n"]
 
         for name, value in extra.items():
-            lines.append(f"{indent}    {name}: {value}\n")
+            lines.append(f"{indent}    {name}: {_safe_str(value)}\n")
 
         return "".join(lines)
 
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):
this is in part ad odds with switching to repr in another pr
also as far as i understood there is a general understanding that str/repr should never fail for object that completed their ctor

RonnyPfannschmidt:
Written by Claude Opus 5.5 via Claude Code for the PR author; I prompted it, it did the work, I read it.

Superseded by #728, which is now merged and carries this PR's contribution. See the "Relation to the existing PRs" section there for what was taken from each. Thank you. Closing.

PR #669, opened 2026-04-29
chore: fix coverage report paths

mschoettle:
Noticed when working on #668 that the coverage report shows missing coverage when it is indeed covered.

Did some troubleshooting and found out that the files are shown inside `.tox` (see: https://github.com/pytest-dev/pluggy/actions/runs/25109365911/job/73578995307?pr=668#step:7:45).

`coverage combine` is required to rewrite the paths.

As a result, there are probably a few `# pragma: no cover` that can be removed.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,9 +7,10 @@ include =
   */pypy*/site-packages/pluggy/*
   *\Lib\site-packages\pluggy\*
 branch = 1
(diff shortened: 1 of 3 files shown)


RonnyPfannschmidt (approved):

RonnyPfannschmidt:
thanks

PR #672, opened 2026-05-06
rewired downstream test drivers

RonnyPfannschmidt:
there was bitrot in the shell scripts
i added a python driver and a workflow to run them on github

testing them now - in particular conda and hatch may need some more details

psa - im also working on actively removing pluggy from hatch as its used in a non-idiomatic way there that doesn’t bring value


RELEASING.rst
```diff
@@ -3,7 +3,8 @@ Release Procedure
 
 #. Dependening on the magnitude of the changes in the release, consider testing
```
RonnyPfannschmidt:
```suggestion
#. Depending on the magnitude of the changes in the release, consider testing
```

RonnyPfannschmidt:
ignoring the hatch errors due to the content

PR #684, opened 2026-06-04
Escape surrogate values in tracing output

### LINKED ISSUE #681: Tracing crashes with UnicodeEncodeError on surrogate escapes in hook arguments/results
## Problem

## Analysis

## Intended Solution

   ```python
   ```

   ```python
   ```

```
      plugin_name: 'lfplugin'
```

```
```

## References

- pytest-dev/pytest#13750

Himanshuagrawal4:
## Summary
- Use 
epr() for hook kwarg values in trace output so surrogate characters are escaped safely.
- Pre-format hook return values before writing tracing output while keeping structural trace labels readable.
- Add regression tests for surrogate values in hook arguments and return values.

## Tests
- git diff --check
- PYTHONPATH=src python -m pytest testing\\test_tracer.py::test_dictargs_escape_surrogate_values testing\\test_pluginmanager.py::test_hook_tracing_escapes_surrogate_values -q (2 passed)
- PYTHONPATH=src python -m pytest testing\\test_tracer.py testing\\test_pluginmanager.py -q (49 passed)

Fixes #681.

### DIFF
--- a/changelog/681.bugfix.rst
+++ b/changelog/681.bugfix.rst
@@ -1,0 +1,1 @@
+Fix tracing output to safely escape surrogate characters in hook arguments and return values.
--- a/src/pluggy/_manager.py
+++ b/src/pluggy/_manager.py
@@ -503,7 +503,7 @@ class PluginManager:
             kwargs: Mapping[str, object],
         ) -> None:
             if outcome.exception is None:
-                hooktrace("finish", hook_name, "-->", outcome.get_result())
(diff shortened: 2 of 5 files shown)


RonnyPfannschmidt:
Written by Claude Opus 5.5 via Claude Code for the PR author; I prompted it, it did the work, I read it.

Superseded by #728, which is now merged and carries this PR's contribution. See the "Relation to the existing PRs" section there for what was taken from each. Thank you. Closing.

PR #694, opened 2026-06-30
unattended ai mess

nyxst4ck:
## Summary
- Roll back plugin manager state when plugin registration fails during hook validation or history replay.
- Add a regression test covering both the plugin name registry and hook implementation cleanup.

## Root cause
`PluginManager.register()` inserted the plugin into `_name2plugin` before validating all hook implementations. If a later hook implementation raised `PluginValidationError`, earlier state remained visible and the failed plugin name could not be reused.

## Tests
- `uv run pytest testing/test_pluginmanager.py::test_register_validation_failure_rolls_back` (failed before the fix, passes after)
- `uv run pytest`

### DIFF
--- a/changelog/694.bugfix.rst
+++ b/changelog/694.bugfix.rst
@@ -1,0 +1,1 @@
+Roll back plugin registration state when a validation error is raised partway through registering a plugin.
--- a/src/pluggy/_manager.py
+++ b/src/pluggy/_manager.py
@@ -151,26 +151,28 @@ class PluginManager:
                 f"{plugin_name}={plugin}\n{self._name2plugin}"
             )
 
-        # XXX if an error happens we should make sure no state has been
-        # changed at point of return
         self._name2plugin[plugin_name] = plugin
 
-        # register matching hook implementations of the plugin
-        for name in dir(plugin):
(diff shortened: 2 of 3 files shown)


RonnyPfannschmidt:
question - where is this used - the usually graceful exit is the followup to a broken plugin at least in the apps im aware of

nyxst4ck:
You are right to ask for a concrete consumer. I searched the local source/docs and found no project call site that catches `PluginValidationError` and then continues registration: `load_setuptools_entrypoints()` propagates it, and the common host behavior is to exit.

The change was prompted by the existing `register()` XXX about returning with changed state, not by a reported downstream recovery failure. The focused test is valid in the project environment (`tox -e py311 -- testing/test_pluginmanager.py -q`: 44 passed), but it only establishes the invariant and not a demonstrated consumer need.

I therefore do not want to overstate this as a bug fix. If maintaining atomic registration is not a desired API invariant, I am happy to close the PR rather than add behavior without a concrete caller.

travis-leith:
"You are right to push back on this" is a classic AI line.

RonnyPfannschmidt:
closing as unattended ai

PR #695, opened 2026-06-30
Handle specname hooks in subset callers

CodingFeng101:
## What changed

`PluginManager.subset_hook_caller()` now builds its exclusion set from the hook implementations registered on the target hook instead of checking whether each plugin has an attribute with the hook name.

## Why

A plugin can implement a hook through `@hookimpl(specname="...")`, so the plugin object does not need to expose an attribute with the hook name. In that case, `subset_hook_caller()` previously failed to exclude the plugin and the implementation still ran.

## Validation

- `PYTHONPATH=src python -m pytest testing\test_pluginmanager.py::test_subset_hook_caller_with_specname testing\test_pluginmanager.py::test_subset_hook_caller -q`
- `PYTHONPATH=src python -m pytest testing\test_pluginmanager.py -q`
- `PYTHONPATH=src python -m pytest -q`
- `uv run pytest -q`
- `uv run pre-commit run -a`

### DIFF
--- a/src/pluggy/_manager.py
+++ b/src/pluggy/_manager.py
@@ -515,7 +515,12 @@ class PluginManager:
         method which manages calls to all registered plugins except the ones
         from remove_plugins."""
         orig: HookCaller = getattr(self.hook, name)
-        plugins_to_remove = {plug for plug in remove_plugins if hasattr(plug, name)}
+        remove_plugins = list(remove_plugins)
+        plugins_to_remove = {
+            hookimpl.plugin
+            for hookimpl in orig.get_hookimpls()
+            if any(hookimpl.plugin is plugin for plugin in remove_plugins)
+        }
         if plugins_to_remove:
             return _SubsetHookCaller(orig, plugins_to_remove)
         return orig
--- a/testing/test_pluginmanager.py
+++ b/testing/test_pluginmanager.py
@@ -497,6 +497,32 @@ def test_subset_hook_caller(pm: PluginManager) -> None:
     assert repr(hc) == "<_SubsetHookCaller 'he_method1'>"
 
 
+def test_subset_hook_caller_with_specname(pm: PluginManager) -> None:
+    class Hooks:
+        @hookspec
+        def he_method1(self, arg):
+            pass
+
+    pm.add_hookspecs(Hooks)
+
+    class Plugin1:
+        @hookimpl(specname="he_method1")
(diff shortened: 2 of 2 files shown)


src/pluggy/_manager.py
```diff
@@ -515,7 +515,12 @@ def subset_hook_caller(
         method which manages calls to all registered plugins except the ones
         from remove_plugins."""
         orig: HookCaller = getattr(self.hook, name)
-        plugins_to_remove = {plug for plug in remove_plugins if hasattr(plug, name)}
+        remove_plugins = list(remove_plugins)
+        plugins_to_remove = {
+            hookimpl.plugin
+            for hookimpl in orig.get_hookimpls()
+            if any(hookimpl.plugin is plugin for plugin in remove_plugins)
+        }
```
bluetech: I reckon the following would be faster, but need to measure:

```suggestion
        plugins_to_remove = set(remove_plugins)
        plugins_to_remove.intersection_update(
            hookimpl.plugin for hookimpl in orig._hookimpls
        )
```

testing/test_pluginmanager.py
```diff
@@ -497,6 +497,32 @@ class PluginNo:
     assert repr(hc) == "<_SubsetHookCaller 'he_method1'>"
 
 
+def test_subset_hook_caller_with_specname(pm: PluginManager) -> None:
+    class Hooks:
+        @hookspec
+        def he_method1(self, arg):
```
bluetech: Let's drop the `arg` since it's inconsequential to this test. For clarity I would just make `Plugin1` return `1` and `Plugin2` return `2`.

bluetech (commented):
This is indeed a bug, thanks for reporting.

`subset_hook_caller` is somewhat performance-sensitive for pytest (used quite a lot, with potentially many plugins registered). So we need to try not to make it too slow.

I left some comments on how to make your approach (checking against all orig's hookimpls) a little faster (though I haven't measured).

Another approach to avoid the problem would be to just drop the `if hasattr(plug, name)` check, which is just an optimization, to avoid the `_SubsetHookCaller` indirection in case it won't remove anything. But that optimization also might be important to pytest, since this is common.

Basically we need to measure the effect both approaches have on pytest and choose the best one...

bluetech (approved):
I updated according to my suggestion, measuring in pytest it seems OK.

PR #702, opened 2026-07-21
Fix surrogate values in hook tracing

### LINKED ISSUE #681: Tracing crashes with UnicodeEncodeError on surrogate escapes in hook arguments/results

almondsun:
## Summary

- render traced hook argument values with `repr` while preserving structural labels
- pre-format traced hook results with `repr` at the `PluginManager` boundary
- cover surrogate-containing kwargs and first-result hook returns
- add a bug-fix changelog fragment

Closes #681.

## Root cause and impact

Tracing formatted arbitrary Python values with `str`, which could leave literal
surrogate characters in the resulting message. Writers using ordinary UTF-8
encoding then raised `UnicodeEncodeError`. Applying `repr` only at the two value
positions safely escapes surrogates without quoting trace labels or double-
representing unrelated structural content.

## Validation

- `uv run pytest testing/test_tracer.py testing/test_pluginmanager.py::test_hook_tracing testing/test_pluginmanager.py::test_hook_tracing_safely_formats_surrogate_result` — 8 passed
- `uv run pytest` — 141 passed
- `uv run pre-commit run -a` — all hooks passed, including Ruff, Flake8, MyPy, formatting, and ReST checks
- `git diff --check`

I used Codex as an implementation assistant, reviewed and validated the complete
change, and take responsibility for the contribution and follow-up.

### DIFF
--- a/changelog/681.bugfix.rst
+++ b/changelog/681.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed tracing crashes when hook arguments or return values contain surrogate
+characters.
--- a/src/pluggy/_manager.py
+++ b/src/pluggy/_manager.py
@@ -503,7 +503,7 @@ class PluginManager:
             kwargs: Mapping[str, object],
         ) -> None:
             if outcome.exception is None:
-                hooktrace("finish", hook_name, "-->", outcome.get_result())
+                hooktrace("finish", hook_name, "-->", repr(outcome.get_result()))
             hooktrace.root.indent -= 1
 
         return self.add_hookcall_monitoring(before, after)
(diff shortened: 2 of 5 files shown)


yangfan-yf-yf (approved):
﻿Reviewed `b7179b9` independently.

The change keeps structural trace labels on `str()` while applying `repr()` only to arbitrary traced values, so surrogate characters are escaped without changing the label format. The two regression tests cover both kwargs and hook-result paths.

Validation on Windows with Python 3.12:

- `python -m pytest` ? 141 passed
- `git diff --check` ? clean

The GitHub test matrix and pre-commit checks are also passing.

Issue #417, opened 2023-07-10
1.2.0: strange output on generate documentation.


kloczek:
Here is sphinx output
```console
+ /usr/bin/sphinx-build -n -T -b man docs build/sphinx/man
Running Sphinx v6.2.1
making output directory... done
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://docs.pytest.org/en/latest/objects.inv...
loading intersphinx inventory from https://setuptools.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://tox.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://devpi.net/docs/devpi/devpi/stable/+doc/objects.inv...
loading intersphinx inventory from https://kedro.readthedocs.io/en/latest/objects.inv...
intersphinx inventory has moved: https://setuptools.readthedocs.io/en/latest/objects.inv -> https://setuptools.pypa.io/en/latest/objects.inv
intersphinx inventory has moved: https://tox.readthedocs.io/en/latest/objects.inv -> https://tox.wiki/en/latest/objects.inv
intersphinx inventory has moved: https://kedro.readthedocs.io/en/latest/objects.inv -> https://docs.kedro.org/en/latest/objects.inv
building [mo]: targets for 0 po files that are out of date
writing output...
building [man]: all manpages
updating environment: [new config] 3 added, 0 changed, 0 removed
reading sources... [100%] index
WARNING: autodoc: failed to import method '_Result.force_exception' from module 'pluggy._callers'; the following exception was raised:
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/sphinx/util/inspect.py", line 341, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: type object '_Result' has no attribute 'force_exception'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/sphinx/ext/autodoc/importer.py", line 106, in import_object
    obj = attrgetter(obj, mangled_name)
  File "/usr/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 324, in get_attr
    return autodoc_attrgetter(self.env.app, obj, name, *defargs)
  File "/usr/lib/python3.8/site-packages/sphinx/ext/autodoc/__init__.py", line 2769, in autodoc_attrgetter
    return safe_getattr(obj, name, *defargs)
[...]

PR #715, opened 2026-07-28
docs: fix nitpicky build for TYPE_CHECKING-only annotations

davidpavlovschi:
## Summary

`tox -e docs` currently fails on `main` because Sphinx runs in nitpicky mode with warnings treated as errors and cannot resolve four private annotation targets from `pluggy._manager`.

The annotations now fall back to their source spellings when `typing.get_type_hints()` encounters names imported only under `TYPE_CHECKING`. The existing `nitpick_ignore` entries cover the qualified spellings, but not the emitted `_tracing.TagTracerSub`, `DistFacade`, and `_Plugin` targets.

This adds those three precise targets to `nitpick_ignore`. It does not broaden the ignore pattern or change runtime code.

Refs pytest-dev/pluggy#417.

## Test plan

- `uv run pytest` — 141 passed
- `uv run pre-commit run -a` — all 11 hooks passed
- `uv run --with tox tox -e docs` — build succeeded (failed before this change)

## AI assistance

I used Claude Code with Claude Opus 5 to investigate the regression, prepare the patch, and run an adversarial review. The commit includes the corresponding co-author and session attribution.

### DIFF
--- a/changelog/417.trivial.rst
+++ b/changelog/417.trivial.rst
@@ -1,0 +1,1 @@
+Fix the ``tox -e docs`` build, which failed under ``sphinx-build -W`` because nitpicky mode reported unresolved references to the private names ``_tracing.TagTracerSub``, ``DistFacade`` and ``_Plugin``.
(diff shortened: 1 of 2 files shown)


bluetech (approved):
Thanks, I ran into this as well.

PR #718, opened 2026-08-04
test: cover pluggy.__version__ getattr and HookCaller._remove_plugin …

darrenhuai:
…not-found path

Ran the suite under coverage.py and found two real gaps in otherwise near-100% covered code. The module-level __getattr__ that lazily resolves pluggy.__version__ via importlib.metadata (added per changelog/590.trivial.rst) had no test exercising either branch. And HookCaller._remove_plugin's ValueError, raised when asked to remove a plugin that never registered on that hook, was untested even though the normal removal path is well covered through
PluginManager.unregister.

### DIFF
--- a/testing/test_details.py
+++ b/testing/test_details.py
@@ -230,6 +230,15 @@ def test_dist_facade_identity_equality_and_hash() -> None:
     assert {fc1: "ok"}[fc1] == "ok"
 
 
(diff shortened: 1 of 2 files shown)


RonnyPfannschmidt (approved):
thanks - the _remove_plugin handling is subject to a more detailed change at a later point

PR #728, opened 2026-09-12
Make hook tracing unable to fail a hook call

### LINKED ISSUE #424: Avoid erroring from tracing

### LINKED ISSUE #681: Tracing crashes with UnicodeEncodeError on surrogate escapes in hook arguments/results

RonnyPfannschmidt:
> **AI-authored.** I asked Claude Code (Opus 5) to work through the tracing PR
> cluster and produce a single change superseding it. The code, the tests, the
> measurements and the text below are the agent's work. I read it and I am
> posting it, and I will follow up on review comments myself.

Closes #424. Closes #681. Supersedes #627, #666, #684, #716.

## The two bugs

Tracing can turn a working hook call into a failing one:

- **#424** — an object whose `__repr__`/`__str__` raises propagates that exception
  out of the hook call (originally [kedro-org/kedro#2630](https://github.com/kedro-org/kedro/issues/2630)).
- **#681** — a lone surrogate in a hook argument or return value produces a message
  the writer cannot encode (originally [pytest-dev/pytest#13750](https://github.com/pytest-dev/pytest/issues/13750)).

Against `main`, with a writer that encodes to utf-8 the way pytest's `--debug` file does:

```
surrogate    UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800' in position 25
broken repr  RuntimeError: repr is broken
```

With this branch, both print `OK`.

## The change

One helper in `_tracing.py`, in the shape of pytest's `saferepr` (as #424 asked for),
applied to both the positional labels and the keyword values:

- `_safe_str()` — `KeyboardInterrupt` and `SystemExit` still propagate, anything else
  renders as `<[RuntimeError('str is broken') raised in str()] BrokenStr object at 0x...>`.
  `BaseException` rather than `Exception`, because pytest's `OutcomeException` — and so
  `pytest.fail()`/`skip()` — sits outside the `Exception` hierarchy, and a `__repr__`
  that fails a test while `--debug` is on is exactly the reported scenario.
- `_escape_surrogates()` afterwards — `backslashreplace`, which also covers surrogates
  that come out of an object's own `__repr__`, a case `repr()` alone does not fix. It
  short-circuits on ASCII text, which is nearly all of it.

**`pytest --debug` output is unchanged**: 674 trace lines on a sample run, byte for byte
identical to `main` once addresses are normalised. Nothing is added, nothing is quoted,
nothing is reformatted — the only difference is that the two inputs above no longer raise.

## Why the trace output is left alone, despite #681

#681 and #627 also called for rendering traced values with `repr()` so their type is
visible (`plugin_name: 'lfplugin'`, `start_path: PosixPath('/x')`). The first commit here
implements that; the second backs it out. Both are kept so the decision is reviewable.

**The reason is that the trace is pytest UX.** `--debug` output is read by a human
scanning for the hook that misbehaved, and a fix for a crash nobody hits in normal use
should not make that scan worse for everyone who does. I am not willing to take that
trade in pluggy — the fix has to be invisible to people who were not hitting the bug.

The worst case is a value meant to be read as a block. With `enable_assertion_pass_hook`,
pytest passes the assertion explanation to `pytest_assertion_pass` as a multi-line string.
Under `str()` the trace shows it as written:

```
    expl: {'x': [0, 1, ...} == {'x': [0, 1, ...}

      Omitting 2 identical items, use -vv to show
      Use -v to get more diff
```

Under `repr()` the same value becomes one escaped line:

```
    expl: "{'x': [0, 1, ...} == {'x': [0, 1, ...}\n  \n  Omitting 2 identical items, use -vv to show\n  Use -v to get more diff"
```

The measurement that settles it, taken on a real `pytest --debug` run — 439 traced kwarg
values, of which 123 render differently under `repr()`:

| what changes | count |
|---|---|
| a string gains quotes (`lfplugin` → `'lfplugin'`) | ~115 |
| a path gains its type (`/tmp/x` → `PosixPath('/tmp/x')` / `local('/tmp/x')`) | 13 |
| an enum gains its name (`1` → `<ExitCode.TESTS_FAILED: 1>`) | 2 |

94% of the delta is quote noise on strings that were already readable. The genuine wins
are real — `path` and `collection_path` printing the same text while being a
`py.path.local` and a `PosixPath` is exactly the confusion a trace should resolve — but
they are 15 lines out of 674, and they are not worth the other 115.

None of that blocks the bug fixes: **they never depended on `repr()`.** The guard and the
surrogate escaping live in `_safe_str`, so #424 and #681 close either way. Surrogates
render as `arg: \ud800` rather than `arg: '\ud800'`; both are legible, neither raises.

So the type-visibility idea is not rejected, only unbundled. If it is still wanted it can
be argued on its own merits in #681 — as a deliberate output change, with its own
changelog entry and its own discussion about what pytest's debug log should look like —
rather than riding along with a crash fix that has to ship regardless.

The one thing `repr()` would buy on such a value is keeping the log's one-line-per-value
structure intact for a machine reading it. That is a parsing concern, and it does not
outweigh making the block unreadable for the person the log is written for.

## Relation to the existing PRs

Each approach against the same matrix (all cells measured, not assumed):

| input | `main` | #666 | #684 | #716 | this PR |
|---|---|---|---|---|---|
| surrogate in a kwarg | crash | crash | fixed | fixed | fixed |
| surrogate in the hook result | crash | crash | fixed | fixed | fixed |
| object whose `__repr__` *returns* a surrogate | crash | crash | crash | fixed | fixed |
| object with a broken `__repr__` | crash | fixed | crash | crash | fixed |
| value passed positionally by a downstream `pm.trace()` caller | crash | fixed | not covered | fixed | fixed |
| `pytest --debug` output unchanged | — | yes | no (16% of lines, plus escaped blocks) | yes | yes |

- **#666**'s guard is carried over and widened — `BaseException` rather than `Exception`,
  and pytest's message format rather than a bespoke one. Its test is carried over. Thanks
  @ShipItAndPray.
- **#716**'s insight that positional args need sanitising too is carried over, and it is
  what keeps downstream `pm.trace()` callers (pytest calls it directly) safe — and what
  makes the `_manager.py` change in #684 unnecessary. Thanks @ltsyk.
- **#684** contributed the regression tests for both the kwarg and the result path, which
  are carried over; its `repr()` change is the part discussed above. Thanks @Himanshuagrawal4.
- **#627** is where all of this was worked out. Thanks @Ghanchu.

## Testing

- `uv run pytest` — 180 passed.
- `uv run pre-commit run -a` — all hooks pass.
- `_tracing.py` at 100% statement and branch coverage.
- 13 new tests: 7 fail on `main`, 2 pin the existing str rendering and the already-correct
  handling of legible non-ASCII, and 4 cover the guards themselves — an exception whose
  own repr fails, one whose repr and str both fail, and `KeyboardInterrupt` raised from
  the value and from the explanation.
- `pytest --debug` output diffed against `main`: 0 differences.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1001,13 +1001,13 @@ Each hook call is traced with its keyword arguments, followed by a ``finish``
 line carrying the result::
 
     he_method1 [hook]
-        arg: 'value'
-        path: PosixPath('/tmp')
+        arg: value
+        path: /tmp
     finish he_method1 --> ['value'] [hook]
 
-Traced values are rendered with :func:`repr`, so their type stays visible, and
-the rendering is defensive: an object whose ``__repr__`` raises is shown as
-``<[RuntimeError(...) raised in repr()] Broken object at 0x...>``, and lone
+Values are rendered with :func:`str`, and the rendering is defensive: an object
+whose ``__str__`` raises is shown as
+``<[RuntimeError(...) raised in str()] Broken object at 0x...>``, and lone
 surrogates are backslash-escaped, so enabling tracing can never turn a working
 hook call into a failing one.
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -997,6 +997,20 @@ undo function to disable the behaviour.
     pm.trace.root.setwriter(print)
     undo = pm.enable_tracing()
 
+Each hook call is traced with its keyword arguments, followed by a ``finish``
+line carrying the result::
+
+    he_method1 [hook]
+        arg: 'value'
+        path: PosixPath('/tmp')
+    finish he_method1 --> ['value'] [hook]
+
+Traced values are rendered with :func:`repr`, so their type stays visible, and
+the rendering is defensive: an object whose ``__repr__`` raises is shown as
+``<[RuntimeError(...) raised in repr()] Broken object at 0x...>``, and lone
+surrogates are backslash-escaped, so enabling tracing can never turn a working
+hook call into a failing one.
+
 
 Call monitoring
 ---------------
--- a/src/pluggy/_tracing.py
+++ b/src/pluggy/_tracing.py
@@ -13,6 +13,60 @@ _Writer = Callable[[str], object]
 _Processor = Callable[[tuple[str, ...], tuple[Any, ...]], object]
 
 
+def _try_repr_or_str(obj: object) -> str:
+    try:
+        return repr(obj)
+    except (KeyboardInterrupt, SystemExit):
+        raise
+    except BaseException:
+        return f'{type(obj).__name__}("{obj}")'
+
+
+def _format_repr_exception(exc: BaseException, obj: object, func: str) -> str:
+    try:
+        exc_info = _try_repr_or_str(exc)
+    except (KeyboardInterrupt, SystemExit):
+        raise
+    except BaseException as inner:
+        exc_info = f"unpresentable exception ({_try_repr_or_str(inner)})"
+    name = type(obj).__name__
+    return f"<[{exc_info} raised in {func}()] {name} object at 0x{id(obj):x}>"
+
+
+def _escape_surrogates(text: str) -> str:
+    """Escape lone surrogates so the result survives any text writer.
+
+    ``repr()`` passes surrogates through unchanged when they originate in an
+    object's own ``__repr__``, and writing such a string to a utf-8 target
+    raises :exc:`UnicodeEncodeError` inside the trace call.
+    """
+    if text.isascii():
+        return text
+    return text.encode("utf-8", "backslashreplace").decode("utf-8")
+
+
+def _safe_str(obj: object) -> str:
+    """``str(obj)`` for structural trace labels, guaranteed not to raise."""
+    try:
+        text = str(obj)
+    except (KeyboardInterrupt, SystemExit):
+        raise
+    except BaseException as exc:
+        text = _format_repr_exception(exc, obj, "str")
+    return _escape_surrogates(text)
+
+
+def _safe_repr(obj: object) -> str:
+    """``repr(obj)`` for traced values, guaranteed not to raise."""
+    try:
+        text = repr(obj)
+    except (KeyboardInterrupt, SystemExit):
+        raise
+    except BaseException as exc:
+        text = _format_repr_exception(exc, obj, "repr")
+    return _escape_surrogates(text)
+
+
 class TagTracer:
     def __init__(self) -> None:
         self._tags2proc: dict[tuple[str, ...], _Processor] = {}
@@ -29,13 +83,13 @@ class TagTracer:
         else:
             extra = {}
 
-        content = " ".join(map(str, args))
+        content = " ".join(map(_safe_str, args))
         indent = "  " * self.indent
 
         lines = [f"{indent}{content} [{':'.join(tags)}]\n"]
 
         for name, value in extra.items():
-            lines.append(f"{indent}    {name}: {value}\n")
+            lines.append(f"{indent}    {name}: {_safe_repr(value)}\n")
 
         return "".join(lines)
 
--- a/src/pluggy/_tracing.py
+++ b/src/pluggy/_tracing.py
@@ -22,7 +22,7 @@ def _try_repr_or_str(obj: object) -> str:
         return f'{type(obj).__name__}("{obj}")'
 
 
-def _format_repr_exception(exc: BaseException, obj: object, func: str) -> str:
+def _format_str_exception(exc: BaseException, obj: object) -> str:
     try:
         exc_info = _try_repr_or_str(exc)
     except (KeyboardInterrupt, SystemExit):
@@ -30,15 +30,15 @@ def _format_repr_exception(exc: BaseException, obj: object, func: str) -> str:
     except BaseException as inner:
         exc_info = f"unpresentable exception ({_try_repr_or_str(inner)})"
     name = type(obj).__name__
-    return f"<[{exc_info} raised in {func}()] {name} object at 0x{id(obj):x}>"
+    return f"<[{exc_info} raised in str()] {name} object at 0x{id(obj):x}>"
 
 
 def _escape_surrogates(text: str) -> str:
     """Escape lone surrogates so the result survives any text writer.
 
-    ``repr()`` passes surrogates through unchanged when they originate in an
-    object's own ``__repr__``, and writing such a string to a utf-8 target
-    raises :exc:`UnicodeEncodeError` inside the trace call.
+    A lone surrogate reaching the writer raises :exc:`UnicodeEncodeError`
+    inside the trace call for any utf-8 target, such as the file behind
+    pytest's ``--debug``.
     """
     if text.isascii():
         return text
@@ -46,24 +46,18 @@ def _escape_surrogates(text: str) -> str:
 
 
 def _safe_str(obj: object) -> str:
-    """``str(obj)`` for structural trace labels, guaranteed not to raise."""
+    """``str(obj)`` for tracing, guaranteed not to raise and always writable.
+
+    Tracing is a debugging aid, so it must never be the reason a hook call
+    fails, and the rendering stays ``str``-based to keep the trace output
+    readable.
+    """
     try:
         text = str(obj)
     except (KeyboardInterrupt, SystemExit):
         raise
     except BaseException as exc:
-        text = _format_repr_exception(exc, obj, "str")
-    return _escape_surrogates(text)
-
-
-def _safe_repr(obj: object) -> str:
-    """``repr(obj)`` for traced values, guaranteed not to raise."""
-    try:
-        text = repr(obj)
-    except (KeyboardInterrupt, SystemExit):
-        raise
-    except BaseException as exc:
-        text = _format_repr_exception(exc, obj, "repr")
+        text = _format_str_exception(exc, obj)
     return _escape_surrogates(text)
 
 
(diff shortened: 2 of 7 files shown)


docs/index.rst
```diff
@@ -1030,6 +1030,20 @@ undo function to disable the behaviour.
...
 
+Each hook call is traced with its keyword arguments, followed by a ``finish``
+line carrying the result::
+
+    he_method1 [hook]
+        arg: 'value'
+        path: PosixPath('/tmp')
+    finish he_method1 --> ['value'] [hook]
+
+Traced values are rendered with :func:`repr`, so their type stays visible, and
```
bluetech, line 1041: I think the part about the defensive is too much info. I would remove everything after "Traced values are rendered with :func:`repr`" and leave the defensiveness as an implementation detail that hopefully no one ever needs to think about anymore :)

src/pluggy/_tracing.py
```diff
@@ -13,6 +13,54 @@
 _Processor = Callable[[tuple[str, ...], tuple[Any, ...]], object]
 
 
+def _try_repr_or_str(obj: object) -> str:
+    try:
+        return repr(obj)
+    except (KeyboardInterrupt, SystemExit):
+        raise
+    except BaseException:
```
bluetech, line 21: Why catch `BaseException` at all? I suggest only to catch `Exception` and just let all base exceptions propagate.

src/pluggy/_tracing.py
```diff
@@ -13,6 +13,54 @@
...
+        return f'{type(obj).__name__}("{obj}")'
```
bluetech, line 22: Does `_format_str_exception` catch exception because the `"{obj}"` can fail?

docs/index.rst
```diff
@@ -1030,6 +1030,20 @@ undo function to disable the behaviour.
     pm.trace.root.setwriter(print)
     undo = pm.enable_tracing()
 
+Each hook call is traced with its keyword arguments, followed by a ``finish``
```
bluetech, line 1033: I think we don't need to document the exact format and such. Looking at the tracing documentation, I'll send my own proposal for these docs shortly.

src/pluggy/_tracing.py
```diff
@@ -13,6 +13,54 @@
...
+        exc_info = _try_repr_or_str(exc)
+    except (KeyboardInterrupt, SystemExit):
+        raise
+    except BaseException as inner:
+        exc_info = f"unpresentable exception ({_try_repr_or_str(inner)})"
+    name = type(obj).__name__
+    return f"<[{exc_info} raised in str()] {name} object at 0x{id(obj):x}>"
+
+
+def _escape_surrogates(text: str) -> str:
```
bluetech, line 36: `backslashreplace` is technically more general than just surrogates errors (I think?). Maybe this shouldn't refer specifically to surrogates and just say "escape"?

src/pluggy/_tracing.py
```diff
@@ -13,6 +13,54 @@
...
+    """Escape lone surrogates so the result survives any text writer.
+
+    A lone surrogate reaching the writer raises :exc:`UnicodeEncodeError`
+    inside the trace call for any utf-8 target, such as the file behind
+    pytest's ``--debug``.
+    """
+    if text.isascii():
```
bluetech, line 43: This is an optimization? Assuming most input is ascii, I guess it's meaningfully faster than the encode+decode?

src/pluggy/_tracing.py
```diff
@@ -13,6 +13,54 @@
...
+        return text
+    return text.encode("utf-8", "backslashreplace").decode("utf-8")
+
+
+def _safe_str(obj: object) -> str:
+    """``str(obj)`` for tracing, guaranteed not to raise and always writable.
+
+    Tracing is a debugging aid, so it must never be the reason a hook call
```
bluetech, line 51: I think it's pretty clear, so would omit this paragraph.

bluetech (approved):
Thanks, LGTM with some comments.

src/pluggy/_tracing.py
RonnyPfannschmidt: people do fun stuff with outcome exceptions

src/pluggy/_tracing.py
RonnyPfannschmidt: Written by Claude Opus 5.5 via Claude Code for the pluggy reviewers; I prompted it, it did the work, I read it.

Keeping `BaseException`: pytest's `OutcomeException` (`pytest.fail()`/`skip()`) is outside `Exception`, and a `__repr__` that calls one of those while `--debug` is on is the scenario from #424. `KeyboardInterrupt` and `SystemExit` still propagate.

src/pluggy/_tracing.py
RonnyPfannschmidt: Written by Claude Opus 5.5 via Claude Code for the pluggy reviewers; I prompted it, it did the work, I read it.

Yes. `_try_repr_or_str` falls back to `str(exc)` through the f-string, and that can raise too. There is now a one-line comment saying so.

src/pluggy/_tracing.py
RonnyPfannschmidt: Written by Claude Opus 5.5 via Claude Code for the pluggy reviewers; I prompted it, it did the work, I read it.

`backslashreplace` is general, but when encoding to utf-8 the only code points that can't be encoded are lone surrogates. So in this call it only ever escapes surrogates, and the name describes exactly what it does. Happy to rename it if you still prefer that.

src/pluggy/_tracing.py
RonnyPfannschmidt: Written by Claude Opus 5.5 via Claude Code for the pluggy reviewers; I prompted it, it did the work, I read it.

Yes, it's an optimization. The agent measured `isascii()` at about 24 ns against about 140 ns for the encode/decode round trip on typical trace values, and nearly every traced value is ASCII. It only runs while tracing is on, so it's small either way, but it's free.

docs/index.rst
RonnyPfannschmidt: Written by Claude Opus 5.5 via Claude Code for the pluggy reviewers; I prompted it, it did the work, I read it.

Done: the docs paragraph is removed in favour of #739, and the docstring paragraph is dropped.

docs/index.rst
RonnyPfannschmidt: Written by Claude Opus 5.5 via Claude Code for the pluggy reviewers; I prompted it, it did the work, I read it.

Done: the docs paragraph is removed in favour of #739, and the docstring paragraph is dropped.

src/pluggy/_tracing.py
RonnyPfannschmidt: Written by Claude Opus 5.5 via Claude Code for the pluggy reviewers; I prompted it, it did the work, I read it.

Done: the docs paragraph is removed in favour of #739, and the docstring paragraph is dropped.

src/pluggy/_tracing.py
bluetech: > and a `__repr__` that calls one of those while --debug is on is the scenario from https://github.com/pytest-dev/pluggy/issues/424.

A `__repr__` which calls `pytest.fail()`/`pytest.skip()` seems *really* obscure. I don't see it mentioned in #424, did I miss anything?

And besides, if some object *does* do this, do we really want to just swallow it? Seems to me better to propagate it.

src/pluggy/_tracing.py
bluetech: Seems to me that the `f'{type(obj).__name__}("{obj}")'` part is a bit too much. A `__repr__` raising is already and edge case, and we then add a fallback to `__str__` and then another edge-case of it failing.

Seems to me we can just try the repr and if it fails go directly to `raised in repr` message.

bluetech:
@RonnyPfannschmidt did you see my two remaining comments? I can send a PR for it if you'd like.

RonnyPfannschmidt:
Yes resolved

bluetech:
Oops, didn't notice. Thanks

PR #731, opened 2026-09-18
fix: reject PluginManager.register(None)

Pitchfork-and-Torch:
## Summary
`None` is the blocked-name sentinel in `_name2plugin` (`set_blocked` stores `None`). `register(None)` therefore silently behaved like `set_blocked(str(id(None)))`:
- `is_blocked(name)` became true
- `is_registered(None)` became true
- `get_plugins()` omitted it

Raise `TypeError("plugin must not be None")` instead.

## Test plan
- [x] `testing/test_pluginmanager.py::test_register_rejects_none` passes


bluetech (commented):
Makes sense, thanks. I added a changelog entry.

PR #732, opened 2026-09-19
Allow hookspecs to declare default values for arguments

### LINKED ISSUE #170: Howto handle hook changes

bluetech:
This is mainly a backward compatibility feature, to allow a hookspec to add new parameters to an existing hook without breaking existing *callers*.

The immediate motivation for this is coming from pytest-dev/pytest#15035, where pytest wants to add an argument to `pytest_deselect` hook but there are many existing external callers.

This feature has been discussed and implemented before:
- #15
- #43
- #170

although I haven't been able to figure out why it was rejected. To me it seems like an intuitive and natural feature, so I decided to put forward another attempt :)

The call overhead when default is not declared is a function call + 2 ifs, which seems OK to me. (Potentially can inline `_apply_defaults` but I still hold hope that cpython will one day do such trivial optimizations itself).

Note: there is a separate matter with defaults declared by *hookimpls*, which I think currently behave in an unintuitive way. There's an issue for it #442. I will try to tackle it next, as it has some interaction with this feature (see `test_does_not_override_hookimpl_default` test in this PR), but it's preexisting.

Fix #170.


RonnyPfannschmidt (approved):
my key nitpick form last time is that i only ever wanted to allow none as sentinel/default in order to prevent some types of drift

but i recon that this might have been too pedantic

bluetech:
@RonnyPfannschmidt restricting to None/sentinel is possible, but I currently think it's not needed. I will discuss it more on #442 and we can add the restriction if we decide it's best.

PR #734, opened 2026-09-21
Fix partial plugin registration on validation errors

### LINKED ISSUE #733: PluginManager.register leaves partial state after validation failure
A failed PluginManager.register() call can leave a plugin partially registered when a later hook implementation fails validation. On main at 9836e545d68690b30accc1eeb4af408f80efa44, a plugin with a valid a_hook and invalid b_hook raises PluginValidationError but remains in the manager and leaves a_hook installed. Expected: registration failure leaves no plugin mapping or hook implementations, so callers can safely fix and retry. The register implementation currently contains a TODO noting that state should not change on failure.

sb123sb123:
Fixes #733

## Cause

PluginManager.register() mutates the plugin map and hook callers incrementally. If a later hook implementation fails validation, the earlier mutations remain. The existing unregister path also used plugin truthiness when removing the name mapping, so a falsey plugin could survive rollback.

## Fix

Make registration transactional: roll back through unregister() for any registration exception, and distinguish a falsey plugin from the None sentinel used for blocked names.

The regression test registers a falsey plugin whose first hook is valid and whose second hook fails validation, then verifies that neither the plugin mapping nor any hook implementation remains.

## Tests

- PYTHONPATH=src python -m pytest testing/test_pluginmanager.py -q — 67 passed
- PYTHONPATH=src python -m pytest -q — 171 passed
- pre-commit run --all-files — passed (Ruff, autoflake, whitespace, pyupgrade, blacken-docs, rst, mypy)
- python -m build --outdir /g/OSS-PR-200/work/evidence/pluggy-atomic-register/build-dist — passed
- git diff --check — passed

## Limitations

The Windows host has Python 3.11.4; uv and tox are unavailable, so the tox environments and the CI matrix entries for Python 3.10, Python 3.14, PyPy, and Ubuntu were not run locally. The package build emitted existing setuptools license deprecation warnings. Host-wide pip check reports unrelated pre-existing dependency conflicts and is not a project failure.

## AI assistance disclosure

This PR was prepared with AI assistance by Codex acting under the sb123sb123 contributor account's instructions. No AI co-author is included. The submitting account takes responsibility for the change and will handle maintainer review and follow-up.


RonnyPfannschmidt:
closing as theres a preexisting pr

